PR #14, opened 2021-08-28
First prototype of gh workflow for PR creation

### LINKED ISSUE #7: Workflow for taking pictures
We need a workflow that, on PR creation, will build the exampleSite and take photos, uploading them as artifacts into the PR – this makes it easy to see if there are changes in the theme as time goes on.

ScrubbyXIV:
Used to fix #7 

This is currently only a draft version of the workflow, using the same OS version as balance-static for consistency.
Main issues observed:

- Had some issues with installing dependencies, mainly postcss and postcss-cli. Might be a local issue.
- When trying to take screenshots, an exception is thrown that puppeteer was unable to launch the browser. Might be that the running OS is missing some packages for chromium.

### DIFF
--- a/.github/workflows/pull-request-creation.yml
+++ b/.github/workflows/pull-request-creation.yml
@@ -43,7 +43,7 @@ jobs:
           devices: iPhone X
 
       - name: Remove unneeded screenshots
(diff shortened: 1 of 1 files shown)


ScrubbyXIV:
Last run of the workflow with limited screenshots: https://github.com/ScrubbyXIV/glam/actions/runs/1183208187

Issue #23, opened 2021-09-01
discord shortcode


nonowazu:
Should have the following options (hopefully named parameters please):

* Code (the server invite code)
* Size (icon, medium, large)
* Color (optional, specifies background color – will default to a discord color if nothing selected)
* Image (optional, replaces the background with an image)

PR #26, opened 2021-09-01
Added shortcodes directory and added a shortcode for a discord invite.

ductapeofdoom:
Adds support for a discord invite shortcode.

Addresses #23.

Currently, the shortcode uses a static image from Discord's branding material as the base text, but that could be easily changed.

Additionally, as noted in the comments, the styling for the 'small', 'medium', and 'large' breakpoints is currently just width but can be styled as needed.


layouts/shortcodes/discordInvite.html
```diff
@@ -0,0 +1,25 @@
+<!--
```
nonowazu, line 1: I think we should replace this with a hugo comment:

```html
{{/* long
multiline comment here */}}
```

layouts/shortcodes/discordInvite.html
```diff
@@ -0,0 +1,25 @@
...
+Shortcode for Discord Invite
+Usage:
+< discordInvite invite_code=<> size=<> color=<> image=<> >
+ -->
+
+<!-- #5865F2 is discord's offical "blurple" color -->
+<!-- Large, medium, small breakpoints are arbitrary. Currently, just using tailwind width classes, but this can be tweaked however needed. -->
+<a {{ with .Get "invite_code" }} href="https://discord.gg/{{ . }}" {{ end }} >
+<img src="/img/Discord-Logo+Wordmark-White.png"
```
nonowazu, line 10: We should probably add this to the theme or pull that image from a cdn... does discord even have a cdn for this sort of thing?

layouts/shortcodes/discordInvite.html
```diff
@@ -0,0 +1,25 @@
...
+{{ if .Get "size" | eq "large"}}
```
nonowazu, line 11: I think you can do:

```html
{{ size := .Get "size" }}
```
And use that as a variable for the below parts.

nonowazu (changes requested):
This is good work, thank you!

layouts/shortcodes/discordInvite.html
ductapeofdoom: I tried just linking to the image on https://discord.com/branding but it appear Discord blocks linking to that that so I think we would have to self host to use it. I mainly used to for filler so whatever we need to change it to is fine.

layouts/shortcodes/discordInvite.html
ductapeofdoom: Updated in latest commit. Still learning hugo syntax so thanks for the recommendation.

layouts/shortcodes/discordInvite.html
ductapeofdoom: Updated in latest commit.

nonowazu (approved):
Thank you for getting this in!

PR #36, opened 2021-09-07
Card component

ductapeofdoom:
image
Initial run at the card component.


layouts/partials/components/card.html
```diff
@@ -0,0 +1,39 @@
+<div class="card">
+	{{ if isset . "ImagePath" }}
+	<div>
+		<img class="w-full" src="{{ .ImagePath }}" />
+	</div>
+	{{ end }}
+	{{ range $element_index, $element_value := .Items}}
+	{{ $color := "#323f4b"}}
+	{{ if eq (mod $element_index 2) 1 }}
```
nonowazu, line 9: I feel like this can be done with some css trickery by adding a class (say `card-item`) and then doing something like:

```css
.card-item:nth-child(even) {
    color: var(--card-even-color);
}

.card-item:nth-child(odd) {
    color: var(--card-odd-color);
}
```

(assuming your talk about making variables on discord)

nonowazu (changes requested):
I think this is great, just 1 discussion point and something I think we can shift to css directly.

layouts/partials/components/card.html
nonowazu: Actually, looking at this again we might make a parent div to contain the card items in for this.

nonowazu (approved):

ductapeofdoom:
Merged as part of #39

PR #37, opened 2021-09-10
More job layouts and barebones example content

nonowazu:
This adds a lot of stuff, but it's still essentially unformatted template-wise. Still, we need the start of this content to be able to build anything, so it's still worthwhile.

### DIFF
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -1,0 +1,6 @@
+{{ define "main" }}
+    {{ range .Params.bis }}
+        <h2><a href="{{ .link }}">{{ .name }}</a></h2>
+        {{ .description | markdownify }}
+    {{ end }}
+{{ end }}
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -1,6 +1,6 @@
 {{ define "main" }}
     {{ range .Params.bis }}
         <h2><a href="{{ .link }}">{{ .name }}</a></h2>
-        {{ .description | markdownify }}
+        {{ .description | .Page.RenderString }}
     {{ end }}
 {{ end }}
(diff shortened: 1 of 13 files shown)


layouts/jobs/bis.html
```diff
@@ -0,0 +1,6 @@
+{{ define "main" }}
+    {{ range .Params.bis }}
+        <h2><a href="{{ .link }}">{{ .name }}</a></h2>
+        {{ .description | markdownify }}
```
ductapeofdoom, line 4: Could potentially user `RenderString` instead of `markdownify` here if we thought we might need custom markdown rendering in the future for this.

layouts/jobs/list.html
```diff
@@ -1,15 +1,53 @@
...
 	<div class="row-span-2">
 		<h1>Character image</h1>
 		<img src="" />
 	</div>
-	<div class="row-span-2"><h1>Job Fundamentals</h1></div>
-	<div class="row-span-2"><h1>Basics</h1></div>
+	<div class="row-span-2">
+		<h1>Job Fundamentals</h1>
+		<div>
+			{{ with .Site.GetPage "leveling-guide" }}
```
ductapeofdoom, line 13: I could be misunderstanding the Hugo documentation ([link for reference](https://gohugo.io/functions/getpage/)) but if the intention is to use a relative path I think we need to use the .Page not the .Site variant of the function.

layouts/jobs/list.html
```diff
@@ -1,15 +1,53 @@
...
+		<div>
+			{{ with .Site.GetPage "basic-guide" }}
+				<h3>Basic Guide (LV 90)</h3>
+				<p>Patch: {{ .Params.patch }}</p>
+				<p>Updated: {{ time.Format "2 Jan, 2006" .Params.lastmod }}</p>
+			{{ end }}
+		</div>
+		<div>
+			{{ with .Site.GetPage "openers" }}
+				<h3>Openers (LV 90)</h3>
```
ductapeofdoom, line 37: Small, but you could make the opener level a parameter instead just for future extendibility.

PR #38, opened 2021-09-11
Questionably better menu nav

nonowazu:
Title. Small change, but getting closer all the time.

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -6,15 +6,35 @@ theme = "glam"
 
 [menu]
   [[menu.main]]
-    identifier = 'home'
-    name = 'home'
-    url = '/'
-    weight = -1
+    identifier = 'jobs'
+    name = 'Combat Jobs'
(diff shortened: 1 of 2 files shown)


exampleSite/config.toml
```diff
@@ -6,15 +6,35 @@ theme = "glam"
...
+    url = '/jobs/'
+    weight = 10
   [[menu.main]]
-    identifier = 'about'
-    name = 'about'
-    url = '/about/'
-    weight = 100
+    identifier = 'other-jobs'
+    name = 'Other Jobs'
+    url = '/jobs/'
```
ductapeofdoom, line 16: Other Jobs links to the same page as Combat Jobs? I'm guessing this is supposed to cover DoH/DoL mostly, will these have their own page?

nonowazu: They will, but I was only implementing this in the exampleSite config only to show how the menu bar will look on the final site without implementing all the requisite content.

PR #41, opened 2021-09-12
Basically a hard-coded footer

nonowazu:
Title. Apologies, but the start of the structure is there!


ductapeofdoom (commented):
Consider moving all the lists' contents to `data`. Since the footer is shared regardless it isn't a huge deal either way and I think there are valid style reasons for not iterating over `data` to build the lists but it might be nice if new types of tools/encounter content is added.

PR #45, opened 2021-09-20
Huge pass at role slider cleanup; home page basic structure and cleanup

nonowazu:
There's a lot to unpack here:

* I made the default text color on the body tag white text
* I cleaned up some of the role slider css so we don't have glaring "white" colors on the page
* I added (empty) partials for the remaining items missing on the homepage
* I broke out the data for the slider into more generic roles
* Example config to write data to those roles added to the exampleSite netlify config.yml


assets/css/main.css
```diff
@@ -7,9 +7,11 @@
  */
 
 body {
+	@apply text-white;
 	background-color: #2c3641;
 }
 
+/* TODO: discuss removal? */
```
ScrubbyXIV, line 14: I'd remove it, yes. No point in keeping it,

data/roles/healers.json
```diff
@@ -0,0 +1,10 @@
+{
+    "title": "healers",
+    "name": "Healers",
+    "order": 2,
+    "color": "#61ce7e",
+    "icon": "/theme-assets/healer.png",
```
ScrubbyXIV, line 6: Suggestion to change file name to "healers.png" for consistency with casters role.

nonowazu: I agree, but probably change to `/theme-assets/icon/role/healer.png` (or perhaps `.svg` if we're lucky?)

ScrubbyXIV (approved):

PR #60, opened 2021-11-05
Refactor "Job Single View" and apply new design

Roxas92:
This is my first time contributing to this project. I'm new to hugo so I might not 100% use its pros.

In this PR I tackled some issues regarding the "Job Single View". I noticed that in the current state it is off the design which has been [posted lately in the discord](https://discord.com/channels/277897135515762698/882335434070433823/896906311235670036). You also wrote multipe times that you wish more partials, so I tried my best to do so.

To explain my thoughts:

- Introduced new global css class `.link` which can be used to have the orange looking link
- Renamed the global css class from `.[size]-job-icon` to `.job-icon-[size]` to fit more into the TailwindCSS terminology
- Added the `role` property to `healers/_index.md` so that we can use these information in the layout to e.g. show different border colors. **(There might be a better solution with Hugo over `data/roles` but I didn't really understand that)**
- Removed the `# [JobTitle]` (Scholar) from the `_index.md` so that the actual content and the title can be displayed seperately without transforming it
- Split the whole `jobs/list.html` in smaller partials
- Added some responsive support
- Added some UI interactions when hovering clickable cards

## Cards
Instead of having one big Card partial I suggest having smaller descriptive ones. I wished Hugo had the concept of Higher Order Component so that we could use one general Card partial in the background, but I didn't really find anything about that. However, I would now suggest the following card types:
- `default` - Can be used for displaying normal content, both with header and footer or without them
- `faq` - A special card for displaying a list of questions in the "content"-area
- `guide` - A card which only consists of the header component, is also clickable
- `link` - A clickable card which can be used to display a link with a description
- `bis_table` - A special card which can be used to display a gearset-list (Note: There were a comment in the original bis_table regarding the list count. I don't see any issue here since we have a fixed amount of columns which will get filled left to right and begin a new row once the column is full)

There are also two "deep partials" to avoid code duplication across the card partials. I decided to prefix them with `_` to make clear that they should not be used outside the card partials.
- `_header`
- `_footer`

___

I could not test this templates with other jobs because I think that this requires to add the jobs over the admin panel.


exampleSite/content/jobs/healers/_index.md
```diff
@@ -3,6 +3,7 @@ menu:
     main:
         name: 'Healers'   
         parent: 'jobs'
+        role: 'healer'
```
nonowazu, line 6: This is for menu items in the headers/footers; you'll probably want to make this a not-nested element.

Also, could you update `exampleSite/static/admin/config.yml` to reflect this change? You can can use the [list widget](https://www.netlifycms.org/docs/widgets/#list) for this.

We use the stuff from here to inform the changes that need to go into the production editor in the `balance-static` repo.

layouts/partials/cards/_header.html
```diff
@@ -0,0 +1,18 @@
...
+  {{ if isset . "icon"}}
+  <div class="mr-8">
+    <img src="{{ .icon }}" />
+  </div>
+  {{ end }}
+  <div>
+    <div class="font-bold text-white text-lg uppercase tracking-widest">
+      {{ .name }}
+    </div>
+      {{ if isset . "patch"}}
```
nonowazu, line 11: for a lot of these if guards, you can probably replace that with a [with](https://gohugo.io/functions/with/) statement, since it will omit the entire block if it isn't found.

layouts/partials/job/header.html
```diff
@@ -0,0 +1,11 @@
+<div class="h-60 xl:h-80">
+  <div class="container mx-auto z-10 relative flex items-center h-full">
+    <div class="flex items-center">
+      <img class="job-icon-lg mr-16 filter drop-shadow-lg" src='{{ print "/theme-assets/jobs/" (lower .Title) ".svg" }}' />
+      <h1 class="text-6xl font-semibold filter drop-shadow-lg">{{ .Title }}</h1>
+    </div>
+  </div>
+  <div class="bg-gray-600 absolute top-0 left-0 w-full -z-1" style="height: 460px;">
+    <img class="object-cover w-full h-full" src='{{ print "/theme-assets/jobs/" (lower .Title) "_background.png" }}' />
```
nonowazu, line 9: Out of curiosity, wouldn't this be better?

```html
<img class="object-cover w-full h-full" src="/theme-assets/jobs/{{ lower .Title }}_background.png" }}' />
```

nonowazu (changes requested):

exampleSite/content/jobs/healers/_index.md
Roxas92: That's a great idea! Do you mean I should add a complete new collection? 🤔 There currently is no entry for the "parent roles", only for scholar.

layouts/partials/job/header.html
Roxas92: Absolutely makes sense! Fixed.

exampleSite/content/jobs/healers/_index.md
nonowazu: In this case, I would create a new section (much like how the scholar "section" is laid out) with some "file(s)" for different roles (in this case, we only want a healer file. Create some default object settings (for the menu items, since we don't want people in the editor to change those!), and then add a role widget.

If you feel like that's too much for this PR (because I do try to add appropriate editor changes as the theme expands in functionality), I can PR it tomorrow for you to get a feel for what I'm thinking about with it.

Roxas92: Yeah I would feel more comfortable for with change if you could do that. Once I understand it I can apply future changes by my own.

nonowazu: Sure thing; in the meantime, change the frontmatter to:

```yml
---
menu:
    main:
        name: 'Healers'   
        parent: 'jobs'
role: 'healer'
---
```

layouts/partials/cards/bis_table.html
```diff
@@ -0,0 +1,41 @@
+{{ $borderClass := "gray-600" }}
+
+{{ if isset . "role"  }}
+  {{ $borderClass = .role }}
+{{ end }}
```
nonowazu, line 5: Sorry, just caught this. I feel like the code could be a bit shorter if you use the [default](https://gohugo.io/functions/default/) functionality in hugo, so perhaps something like this?

```
{{ $borderClass := default "gray-600" .role }}
```

Or some such.

nonowazu (changes requested):

layouts/partials/cards/bis_table.html
Roxas92: Ah nice, didn't know about this default thing. Works great!

layouts/partials/cards/_footer.html
```diff
@@ -0,0 +1,11 @@
+{{ if isset . "readMoreLink" }}
+<div class="border-t border-card-dark pt-5 text-right">
+  <a href="{{ .readMoreLink }}" class="link">
+    {{ $readMoreText := "Read more" }}
+    {{ if isset . "readMoreText"}}
+      {{ $readMoreText = .readMoreText }}
+    {{ end }}
+    {{ $readMoreText }} »
```
nonowazu, line 8: Ahh, I'm so sorry, but I missed this too. You can probably simplify this whole section to

```html
{{ default "Read more" .readMoreText }} »
```

On the one hand it's nitpicking, on the other hand, it's *80% line saving*, which makes me feel like [Skeletor when he saves money](https://www.youtube.com/watch?v=nqhLn76kCv0)

layouts/partials/cards/bis_table.html
nonowazu: Uh, I think the `bis_table.html` template is still missing the `default` statement

nonowazu (changes requested):
Okay, I swear I'm going to stop nagging, but two very minor changes left!

nonowazu (approved):
Looks good, thank you for your patience with me in this review process!

PR #61, opened 2021-11-05
Attempts to fix the screenshot PR workflow

nonowazu:
It's sort of hard to explain the underlying nuance that is hugo whining, but I believe it's because it's looking for a local node_modules directory.


nonowazu:
I'm going to close this and submit another PR with a more succinct fix.

PR #63, opened 2021-11-05
Home Page Styling, Layout, Paritals

OHaiDanny:
homepageNov4

In This PR I've updated the homepage to look like what is shown in the mockups

- Added some logic to use a unique header for the homepage only.
- Added partials and data for the encounter section of the homepage
- Added content for the role slider and added some JS to add a unique classname for styling
- lots of tailwind classes to style the page

Let me know if theres any questions or comments :)


assets/js/role_slider.js
```diff
@@ -0,0 +1,12 @@
...
+roleSliderTabs[0].children[0].classList.add('selected');
+
+roleSliderContainer.addEventListener('mousedown', (e) => {
+    roleSliderTabs.forEach(tab => {
+        tab.children[0].classList.remove('selected');
+        if (tab.contains(e.target)) {
+            tab.children[0].classList.add('selected');
+        };
+    });
+});
```
nonowazu, line 12: Out of curiosity, why the JS? From what I remember, the role slider needed no js for either theming or operation.

data/encounter/savage.json
```diff
@@ -0,0 +1,28 @@
+{
```
nonowazu, line 1: Could you add a way to generate encounter data in `exampleSite/static/admin/config.yml`? You can find docs for this [here](https://www.netlifycms.org/docs/intro/)

data/encounter/savage.json
```diff
@@ -0,0 +1,28 @@
...
+    "names": [
+        "edenPromise",
+        "edenVerse",
+        "edenGate",
+        "oldTiers"
+    ],
```
nonowazu, line 7: I *think* these might not be necessary, since I think you can do the following in hugo, but we should probably brainstorm on using `range` with hugo on data to make this auto-iterate. I'll take a look a bit later, but if this ends up being required, then it's required.

data/frontpage/slider.json
```diff
@@ -24,7 +24,7 @@
                 "icon": "/theme-assets/paladin.png"
             },
             {
-                "name": "Bunbreaker",
+                "name": "Gunbreaker",
```
nonowazu, line 27: Goodbye, sweet joke.

layouts/_default/baseof.html
```diff
@@ -2,8 +2,14 @@
...
     <body>
-        {{- partial "header.html" . -}}
-        <div id="content">
+        {{ $title := .Title }}
+        {{ if eq $title "Home"}}
+          {{- partial "header_home.html" . -}}
+        {{ end }}
+        {{ if ne $title "Home"}}
+          {{- partial "header.html" . -}}
+        {{ end}}
```
nonowazu, line 11: You can do an [if..else](https://gohugo.io/templates/introduction/#example-4-if--else) in hugo for this whole part

layouts/index.html
```diff
@@ -2,7 +2,10 @@
 <div class="container mx-auto">
   {{ partial "homepage/role_slider.html" . }}
   {{ partial "homepage/recents.html" . }}
```
nonowazu, line 4: While we're going through this PR, I might ask you to hold off on some of these partials – get that the home page looks empty without some of these, but at the same time we've also got to focus on getting the theme good enough to move over to the primary domain/repo.

layouts/partials/homepage/encounters/card.html
```diff
@@ -0,0 +1,8 @@
+<div class="flex-1 flex-col w-1/4 mr-4 bg-border-dark-color">
+  <a class="block" href="{{ .link_to }}">
+    <img src="{{ .image_link }}" alt="" />
+    <div class="bg-border-dark-color">
+      <h2 class="text-2xl uppercase p-4 font-bold">{{ .name }}</h2>
+    </div>
+  </a>
+</div>
```
nonowazu, line 8: I feel like some of this conflicts with the other active PR right now – you might want to discuss how to handle this with them. They technically submitted their PR first so I'm going to give them first dibs at merging, which might mess up your flow with card partials.

layouts/partials/homepage/role_slider.html
```diff
@@ -1,11 +1,19 @@
-<div class="tabbed">
+<div id="roleSlider" class="tabbed bg-card-dark-background-color">
```
nonowazu, line 1: I might ask that you go into the css file and just `@apply` the changes to the tabbed class instead of adding them here.

layouts/partials/homepage/role_slider/content.html
```diff
@@ -1,12 +1,15 @@
-<div class="tab-content">
+<div class="tab-content p-8">
```
nonowazu, line 1: ditto on `@apply`ing this in the `.tab-content` selector in the main css.

static/theme-assets/icons/role/role_melee.svg
```diff
@@ -0,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="40" height="40" viewBox="0 0 40 40">
+  <g id="Icon" transform="translate(-118)">
+    <g id="Melee2" transform="translate(124.98 5.644)">
+      <path id="Union_7" data-name="Union 7" d="M13.624,27.392l-1.137-4.865,4.791-4.989,4.633-.363,4.336,2.559v5.776l-1.122,1.883ZM.471,19.1,1.1,14.706,5.1,18.551v.975l.707.974.864-1.949L.55,12.705,0,6.368H2.67l.314-2.159L6.2,4.131l.157-2.182h3.3L10.6,0l5.811,2.572,5.5,13.64-5.34.234-5.1,5.5-7.892.582Z" fill="#f15c5c"/>
```
nonowazu, line 4: The second line is drawing incorrectly here.

tailwind.config.js
```diff
@@ -3,14 +3,19 @@ module.exports = {
...
+        'card-dark-background-color' : '#19191a',
         'card-even-color' : '#2a3641',
         'card-border-color' : '#3871c2',
         'card-header-text-color' : '#7b8794',
-        'dark-block-color' : '#1F2933'
+        'dark-block-color' : '#1F2933',
+        'border-dark-color' : "#141415"
       }
     },
   },
```
nonowazu, line 21: Ditto on coordinating with Roxas re: changes here – I'd rather have both of you be happy than me be miserable sorting it out :-)

nonowazu (changes requested):

assets/js/role_slider.js
OHaiDanny: Ah yes, so I noticed that in the mockups the "selected" tab has the same background image as the body of the role slider. I think this is to visually show to the user which tab they have selected when they click around them. I was trying to figure out if there was a way using hugo to add this "selected" class name.

My knowledge on hugo is very basic so I thought maybe it would be easier to just some JS to add that class instead.

data/encounter/savage.json
OHaiDanny: Sounds good :)

data/frontpage/slider.json
OHaiDanny: aww 😢

tailwind.config.js
Roxas92: @OHaiDanny once my PR is merged I can take care of these colors in your PR as well. We most likely used the same color codes but with different naming. I'm more a fan of not having words like "color" inside of color-variables since writing them unser color imply that they serve as color suffixes.

layouts/partials/homepage/encounters/card.html
Roxas92: @OHaiDanny same as in my other comment: If you want to I can check this partial once my PR is merged.

assets/js/role_slider.js
nonowazu: Oh, I see what I missed: if you'll look [here](https://github.com/The-Balance-FFXIV/glam/blob/9a716f4ae94b847536235643b977c2c6a60da74d/assets/css/main.css#L95), I commented out the styling for tabs which are "checked". With that, you should be able to remove all the js and the tabs become a css-only affair (which I'll fully admit I was needlessly proud of).

Try uncommenting out that section and moving your styles there?

layouts/partials/homepage/role_slider.html
OHaiDanny: addressed!

layouts/_default/baseof.html
OHaiDanny: addressed in upcoming code push

layouts/index.html
OHaiDanny: Sounds good. I believe I'll comment out the "Latest Guides" and the "Community Spotlight" sections as that is what the consensus is from the dev channel :)

layouts/partials/homepage/role_slider/content.html
OHaiDanny: updated!

tailwind.config.js
OHaiDanny: I've updated the colors to use what @Roxas92 merged in :)

assets/js/role_slider.js
OHaiDanny: Will have an update in the next PR!

data/encounter/savage.json
OHaiDanny: @nonowazu I made a refactor of the data structure here to accomodate the added code in the `admin/config.yml`. Please let me know if I missed anything (first time doing the config.yml stuff)

assets/js/role_slider.js
OHaiDanny: This has been updated in the latest PR!

data/encounter/savage.json
OHaiDanny: Updated in commit https://github.com/The-Balance-FFXIV/glam/pull/63/commits/6f3ed701a95c5a1704ecec6552bb2c2833a01974

exampleSite/static/admin/config.yml
```diff
@@ -48,6 +48,24 @@ collections:
...
+      - {label: "Category", name: "category", widget: "string"}
+      - label: "Encounters" 
+        name: "encounters"
+        widget: "list"
+        fields:
+          - {label: "Identifier", name: "id", widget: "string"}
+          - {label: "Name of Encounter", name: "name", widget: "string"}
+          - {label: "Image Link", name: "image_link", widget: "string"}
+          - {label: "Encounter Link", name: "encounter_link", widget: "string"}
+    }
```
nonowazu, line 68: I'm seeing this when I use hugo to render and then go to http://localhost:1313/admin/#/:

Screen Shot 2021-11-05 at 6 17 37 PM

I believe it's because you're doing json-like `{}` after the `fields:` view; maybe something more like this?

```yaml
fields:
  - {label: "Identifier", name: "id", widget: "string"}
  - {label: "Name of Encounter", name: "name", widget: "string"}
  - {label: "Image Link", name: "image_link", widget: "string"}
  - {label: "Encounter Link", name: "encounter_link", widget: "string"}
  ```
I might also switch the string widget out for `select` widget (docs [here](https://www.netlifycms.org/docs/widgets/#select)) – this way, we can narrow the category of file(s).

tl;dr - yaml is weird and we should probably make a PR checker that'll yamllint this file just like we have on the production repo.

File: assets/js/role_slider.js [hunk 13]
```diff
@@ -0,0 +1,12 @@
...
+// roleSliderTabs[0].children[0].classList.add('selected');
+
+// roleSliderContainer.addEventListener('mousedown', (e) => {
+//     roleSliderTabs.forEach(tab => {
+//         tab.children[0].classList.remove('selected');
+//         if (tab.contains(e.target)) {
+//             tab.children[0].classList.add('selected');
+//         };
+//     });
+// });
```
nonowazu, line 12: One last nit on this, I swear!

If the css-only setup worked, would you mind removing this file (and from the associated head partial)?

nonowazu (changes requested):
Just 2 small changes and I think it's good to go!

assets/js/role_slider.js
OHaiDanny: Oh shoot I forgot to remove this! My bad!

exampleSite/static/admin/config.yml
OHaiDanny: ahhh my bad. You're correct; the `fields:` shouldn't start with `{` oops! I'll try to test this on the admin page before I push the change.

OHaiDanny: @nonowazu fixed in commit https://github.com/The-Balance-FFXIV/glam/pull/63/commits/f9fbce0568dc5a840fde1d8445d24742fc61b1d3

assets/js/role_slider.js
OHaiDanny: @nonowazu fixed in commit https://github.com/The-Balance-FFXIV/glam/pull/63/commits/f9fbce0568dc5a840fde1d8445d24742fc61b1d3

nonowazu (approved):

PR #64, opened 2021-11-05
Basic netlify implementation for role landing page

nonowazu:
More fields are likely needed, but it should satisfy the role landing page for now (even though role landing page proper is post-soft-launch)

### DIFF
--- a/exampleSite/content/jobs/healers/_index.md
+++ b/exampleSite/content/jobs/healers/_index.md
@@ -1,9 +1,8 @@
 ---
 menu:
-    main:
-        name: 'Healers'   
-        parent: 'jobs'
-role: 'healer'
+  main:
+    name: Healers
+    parent: jobs
+role: healer
 ---
-
-# About healers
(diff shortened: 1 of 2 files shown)


Roxas92:
Looks good to me. Thank you for sharing how to approach those changes in the future!

___

What came in my mind, but I can also tackle this when working on it frontendwise:

Would it make sense to add two new fields for the name in singular and plural? 🤔  Looking at the design for the role page we use both "Healer" and "Healers". Later on for Melee DPS we use "Melee DPS" for both display cases so we can't use some logic which adds an "s". Or do you think we can solve that in a different way?

holl0wstar:
We could either add a new field or stick to a single word for all of them for consistency. 

i.e we could just use a logic to pluralize and instead of saying 'Melee DPS' we use 'Melee(s)', as well.

Thoughts, Nono?

lailleo:
Seems like as good a plan as any, as I don't see any other phrasing of the term that works well.

Issue #68, opened 2021-11-06
Configuration in editor to add configurable pictures to frontmatter


nonowazu:
As an example:

![image](https://user-images.githubusercontent.com/83298827/140621341-4c517bcd-b3d9-4084-b20b-cbddba4959de.png)

And also have a configurable space to dump pictures (like with the [user profile editor](https://github.com/The-Balance-FFXIV/glam/blob/fe6caaa3a08308af22c3ab57a378baacd55e707b/exampleSite/static/admin/config.yml#L59))

nonowazu:
I'm pretty sure #72 addresses this

PR #72, opened 2021-11-07
Proposal: Teaser images for guides / single pages

Roxas92:
As pointed out in https://github.com/The-Balance-FFXIV/glam/issues/68 it makes sense to have configurable images for some parts of the landingpage. I wanted to make a proposal on how to tackle this issue:

I would add a new optional field to _all_ guide / single pages which allows the content editors to upload a teaser picture for this specific page. Whenver this page gets displayed on a page as a "link to the page" we can optionally show a teaser image for it. If frontendwise a image is required we always show an image - we use a fallback of no image is set. We can also - frontendwise - define the size of the image and it then gets fit into this area by either zooming it in or out.

I think with this approach we should be future proof for any kind of displaying logic / crosslinking.

> And also have a configurable space to dump pictures

I would give all pages the same access to `/img/jobs` so if needed images can be shared across the jobs.

## Admin view
image

## Frontend with image
image

## Frontend with no images (but frontend requires those)
image


nonowazu:
While I don't mind this PR in theory, do you intend for these "teaser" images to also be part of the opengraph display? If so, we may want to leverage the internal hugo [opengraph](https://gohugo.io/templates/internal/#use-the-open-graph-template) template (code [here](https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/opengraph.html)).

If that's the case, we'll want to have the "teaser" (highlight?) image be a list with a single image (netlify example [here](https://github.com/xivresources/glossary.xivresources.com/blob/main/static/admin/config.yml#L33-L36)). We might also want to expand out while we're here and set up the other image types for the landing page – or at least for the 4 prominent ones on the pages for now.

Roxas92:
That's actually a good point. I often forget about those cases. I will check this out and come back with an idea how to solve that.

Roxas92:
Hm ... Although I made a commit one hour ago with some changes replacing "teaser" with a list of "opengraph images", but I'm somehow not satisfied with this solution. Maybe your last sentence caused some confusion to me:

> We might also want to expand out while we're here and set up the other image types for the landing page – or at least for the 4 prominent ones on the pages for now.

Does that mean you plan to have multiple teaser/cover/thumbnail (however we want to call them) images for different use cases? If so, my last commit might not work for that since we now have a stupid list of images with no context. If not we cam definitely go with my solution.

Another approach would be to go back to my initital approach but make a custom `opengraph.html`.

But in general, I think I did not 100% get what you want to achieve with that, so I need some more clarity. (Might also be of my bad english skills 🙈 )

nonowazu:
In this case I should have been more clear (but I was tired, so I'll lean on that excuse): Right now in the mockup, there are 4 images on the role landing page:

1. Above the leveling guide
2. Above the basic guide
3. Above the advanced guide
4. Above fight tips

We need to make those configurable in the editor, as they might change over time (that and we don't want to hold content or art asset selection hostage to dev changes).

Thinking about it, I agree that we need to keep a separate image for "highlights" / opengraph display, but for soft launch  let's hold off on that functionality since we really need to get these usable to people clicking on the link and not focus on polish like how the link appears in facebook/twitter/discord/etc.

Roxas92:
Ah, ok I see. Then there might be a misunderstanding.

The solution I brought up works as following: When using the `cards/guide.html` partial, we can optionally define an image which gets displayed at the top of this card. For the 4 cases you mentioned we take the teaser image of the page (guide) and display it as the cards image. So logically they are not just "images on the landingpage" they should be treated as "guide card where we want to display the pages image" - does that make sense?

nonowazu:
That makes sense, but 'teaser' doesn't sound like the right word to me, which I think is what's causing my confusion? When I think of the word "teaser" based on the earlier conversation, I assumed this was preview for opengraph things.

Maybe we define it as the 'card header image' or some such?

Roxas92:
Sounds good to me. So do you want me to go back to the original version with "teasers" and just rename the properties? Meaning to skip this opengraph thing for now and handle this later.

nonowazu:
Yeah, let's go back to the pre-opengraph stuff as it was my misunderstanding to begin with. Should we call them perhaps something like this?

* `leveling_image`
* `basics_image`
* `advanced_image`
* `tips_image`

???

Roxas92:
Like discussed in discord, I reverted the changes for "Opengraph Images" and went back to the "teaser image" solution but with the naming of `card_header_image` to make the usage more clear.

nonowazu (approved):

nonowazu:
This doesn't seem to address the landing page (e.g., `/jobs/healers/scholar/`), but it does fix something missing with the guide pages so I'll push it through.

Issue #67, opened 2021-11-06
Site param-configurable limit to number of bis entries


nonowazu:
Without a limit, things start looking a little... heavy:

![image](https://user-images.githubusercontent.com/83298827/140621279-26e69cf4-e866-4e6d-957a-e017db0848ee.png)

Probably best to use hugo's [first](https://gohugo.io/functions/first/) feature to configure, and use [site parameters](https://gohugo.io/variables/site/#the-siteparams-variable) to configure that centrally.

PR #73, opened 2021-11-07
add global limit for bis table entries

Roxas92:
Adresses https://github.com/The-Balance-FFXIV/glam/issues/67

### DIFF
--- a/layouts/partials/cards/bis_table.html
+++ b/layouts/partials/cards/bis_table.html
@@ -15,18 +15,18 @@
           <div class="col-span-2">
               <div class="text-white font-bold py-3 uppercase tracking-widest">Current Tier</div>
               <div class="grid grid-cols-2 gap-3">
-                  {{ range first site.Params.maxTierGearsets .bis }}
-                      {{ $bisItemClass := "bis-item" }}
-                      {{/*  TODO: figure out a better id than the name. Its too long and has to many spaces to be a clean link.  */}}
-                      <div class="link"><a href={{ print "best-in-slot/#" .name }}>{{ .name }} »</a></div>
+                  {{ $gearsetsList := first site.Params.maxTierGearsets .bis }}
+                  {{ range $index, $item := $gearsetsList }}
+                      <div class="link"><a href={{ printf "best-in-slot/#%d" $index }}>{{ $item.name }} »</a></div>
                   {{ end }}
               </div>
           </div>
           <div>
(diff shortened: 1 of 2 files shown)


layouts/partials/cards/bis_table.html
```diff
@@ -15,19 +15,19 @@
           <div class="col-span-2">
               <div class="text-white font-bold py-3 uppercase tracking-widest">Current Tier</div>
               <div class="grid grid-cols-2 gap-3">
-                  {{ range $bisIndex, $bisName := .bis }}
+                  {{ range first site.Params.maxTierGearsets .bis }}
                       {{ $bisItemClass := "bis-item" }}
                       {{/*  TODO: figure out a better id than the name. Its too long and has to many spaces to be a clean link.  */}}
-                      <div class="link"><a href={{ print "best-in-slot/#" $bisName.name }}>{{ $bisName.name }} »</a></div>
+                      <div class="link"><a href={{ print "best-in-slot/#" .name }}>{{ .name }} »</a></div>
```
nonowazu, line 21: Ha, I think I actually caught a bug from the previous setup. I think this was supposed to be:

```html
<div class="link"><a href="{{ printf "best-in-slot/#%d" $bisIndex }}">{{ $bisName.name }} »</a></div>
```

So the href would link correctly and push over to the right gearset. So you might need to redo this a bit to get the index – sorry!

layouts/partials/cards/bis_table.html
```diff
@@ -15,19 +15,19 @@
...
               </div>
           </div>
           <div>
               <div class="text-white font-bold py-3 uppercase tracking-widest">Ultimates</div>
               <div class="grid grid-cols-1 gap-3">
-                  {{ range $bisName := .bis }}
-                      <div class="link"><a href={{ print "best-in-slot/#" $bisName.name }}>{{ $bisName.name }} »</a></div>
-                  {{ end }}
+                {{ range first site.Params.maxUltimateGearsets .bis }}
+                  <div class="link"><a href={{ print "best-in-slot/#" .name }}>{{ .name }} »</a></div>
```
nonowazu, line 29: Same here with needing the index, sorry again!

nonowazu (changes requested):

layouts/partials/cards/bis_table.html
Roxas92: Alright, maked sense, is fixed!

nonowazu (approved):

PR #76, opened 2021-11-07
Mobile Responsiveness for Homepage

OHaiDanny:
A lot of the code changes here are mainly style and tailwind changes to make the homepage responsive. There are a couple of moderate sized changes in here:

- I was able to restructure the role slider so that when it is in mobile it will behave like an accordion. Once it is in tablet view it will behave like tabs.
- I made the navigation menu on mobile fixed so users are able to access it when they are scrolling down.


assets/js/role_slider.js
```diff
@@ -0,0 +1,12 @@
...
+// roleSliderTabs[0].children[0].classList.add('selected');
+
+// roleSliderContainer.addEventListener('mousedown', (e) => {
+//     roleSliderTabs.forEach(tab => {
+//         tab.children[0].classList.remove('selected');
+//         if (tab.contains(e.target)) {
+//             tab.children[0].classList.add('selected');
+//         };
+//     });
+// });
```
nonowazu, line 12: Weird question – I thought we removed this?

nonowazu (changes requested):
I need to more fully look over this, but I'm also hesitant to spend time on this since mobile isn't a priority for soft launch. Let me mull over this a bit.

assets/js/role_slider.js
OHaiDanny: lol omg my rebasing was acting up on this branch. I'll remove it in a sec 😆

nonowazu:
I really don't want to move forward with this PR for soft launch, but @lailleo tells me that there's a bug that frontend wants fixed tied up as part of this PR - is there a reason it can't be broken out into it's own PR?

OHaiDanny:
> I really don't want to move forward with this PR for soft launch, but @lailleo tells me that there's a bug that frontend wants fixed tied up as part of this PR - is there a reason it can't be broken out into it's own PR?

@nonowazu the bug fix is moved to https://github.com/The-Balance-FFXIV/glam/pull/78

PR #77, opened 2021-11-07
fix shadows and some heading stylings

Roxas92:
This is what I previously prepared in https://github.com/The-Balance-FFXIV/glam/pull/71 but now put into a seperate PR to both have smaller chunks of changes and not interrupt the review process. In theory this is not critical for soft launch but since I already finished it, I would appreciate adding this anyway.

## Extended `.role-header-[role]` class
- I noticed that we use different shadows depending on the selected role. To relfect that in addition to the `.role-header` class (for normal, black shadow) we also have a `.role-header-[role]` class for each of our 3 roles.
- To achieve that we also have now one dropShadow class for each role `.drop-shadow-lg-[role]`

## Misc
- Added "Kumbh Sans" font and made it available over the `.font-head` class
- Updated some margins and paddings

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -3,7 +3,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-<link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css2?family=Kumbh+Sans:wght@700&family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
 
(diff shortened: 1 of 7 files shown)


nonowazu (approved):

layouts/partials/head.html
```diff
@@ -3,7 +3,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-<link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css2?family=Kumbh+Sans:wght@700&family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
```
OHaiDanny, line 6: @Roxas92 Just an FYI here that It looks like we still need Roboto font for body of content. I have it fixed up in my PR :)

PR #78, opened 2021-11-07
Kumbh Sans Font + Role Slider desktop CSS

OHaiDanny:
In this PR I've added the implementation of the Kumbh Sans Font.

I've also made some desktop only adjustment to the role slider with the help of Moth.


nonowazu:
While I appreciate this PR, I really want to focus on landing#72, and making sure guide pages are functional; as soon as those are done, I'll get to this PR. Mobile PR I'll need to push off farther.

layouts/partials/head.html
```diff
@@ -3,7 +3,8 @@
 <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-<link href="https://fonts.googleapis.com/css2?family=Kumbh+Sans:wght@700&family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
```
Roxas92, line 6: You can still use my implementation. It's better to have only one api call to google instead of two.

assets/css/main.css
```diff
@@ -7,7 +7,7 @@
  */
 
 body {
-	@apply text-white;
+	@apply font-kumbhsans text-white;
```
Roxas92, line 10: There is no need to add this globally to body. We should call kumbhsans only when needed.

tailwind.config.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   darkMode: false, // or 'media' or 'class'
   theme: {
     fontFamily: {
+      'kumbhsans': ['Kumbh Sans', 'sans-serif'],
```
Roxas92, line 8: Here we should only use one of them `kumbhsans` or `head` since both have the same content. I'd recommend using `head` since it's not bound the it's actual font and can in the future be replaced without changing the variable name.

tailwind.config.js
```diff
@@ -37,6 +38,7 @@ module.exports = {
       },
       fontSize: {
         'base': '0.9375rem',
+        'content': "15px"
```
Roxas92, line 41: `0.9375rem` are `15px`. So you can skip `content` here. Also, `base` will be applied globally as default to all elements.

Roxas92 (changes requested):

Roxas92:
Looks good to me now! :)

Roxas92 (approved):

OHaiDanny:
Ship it! 😄

OHaiDanny:
Just doing a rebase of `main` branch

nonowazu (approved):

PR #88, opened 2021-11-10
Hide savage and ultimate encounters on homepage

### LINKED ISSUE #85: hide savage/ultimate encounter guides on homepage for soft launch

ScrubbyXIV:
Resolves #85 

Added underscore to encounter key in json to hide guides. This should make reverting this change easier.

Example of home page with this change:
image


ScrubbyXIV:
@nonowazu there's also the option to hide the encounters partial on the homepage. Let me know if you'd prefer that over the PR implementation.

OHaiDanny:
Looks good!

PR #89, opened 2021-11-10
Setup dynamic netlify cms config and add some build commands

Roxas92:
This PR allows us to get rid of the 4.2k lines in our `admin/config.yml` and makes it also easier for uns in the future to maintain this file.

All of these changes **should** be part of https://github.com/The-Balance-FFXIV/balance-static and not of the theme itself as the theme should not be responsible for how the admin cp is set up. So once everyone is cool with this, I would also add those changes to https://github.com/The-Balance-FFXIV/balance-static and let them in the theme only for testing purposes.

I also extended the `README` so that you know how the test the changes.

Hope it works out for you, looking forward to get your feedback on that!


Roxas92:
FYI I plan some more things for this PR, so you can still test it but be prepared for some changes. Following changes are planned:
- `job_name` for Job Landingpages to replace the current Title solution
- Add the 4 images from the landingpage to the netlify config

cms/collections/jobs.js
```diff
@@ -0,0 +1,389 @@
...
+    createThumbnailField(),
+    createAuthorsField(),
+    createPatchField(),
+    createLastUpdatedField(),
+    createChangelogField(),
+  ],
+});
+
+const createFAQ = (job, folderPrefix) => ({
+  label: 'Frequently Asked Questions',
```
OHaiDanny, line 215: I'm wondering with the recent discussions of https://github.com/The-Balance-FFXIV/balance-static/issues/589 do we need to add the `layout: 'qna'` into these? My guess is that if we don't then the layout will default to `single.html`?

Just want to future proof any new pages that content creators make

cms/collections/jobs.js
```diff
@@ -0,0 +1,389 @@
...
+    createThumbnailField(),
+    createAuthorsField(),
+    createPatchField(),
+    createLastUpdatedField(),
+    createChangelogField(),
+  ],
+});
+
+const createBestInSlot = (job, folderPrefix) => ({
+  label: 'Best in Slot',
```
OHaiDanny, line 261: Same thing with my comment on FAQ. Do we need to hard code `layout 'bis'`?

cms/collections/jobs.js
```diff
@@ -0,0 +1,389 @@
...
+    {
+      label: 'Priority',
+      name: 'priority',
+      widget: 'string',
+    },
+  ],
+});
+
+const createJobChanges = (job, folderPrefix) => ({
+  label: 'Job Changes',
```
OHaiDanny, line 327: Same thing with my comment on FAQ. Do we need to add `layout:'changes'`

cms/collections/jobs.js
Roxas92: Discussed that in Discord. No Changes needed.

Roxas92:
To summarize the latest changes;
- Adjust the Role Names to fit into the folder structure and the rest
- Add some possibilities to add images, like on the job landingpage
- Generally applied some fixes to make the new admin config work with the code

____

I know these a lots of changes but it would have been hard to split them into multipe PRs since they all depend on the new netlify config feature.

nonowazu:
I  notice the github workflow breaks here, talking about node stuff – could you look into that? I want to make sure this works smoothly when we start using this theme with the content repo.

README.md
```diff
@@ -2,14 +2,31 @@
...
 
+## Install Hugo
+Follow the instruction on [gohugo.io](https://gohugo.io/getting-started/installing/) to install hugo on your machine.
+
 ## Testing the theme locally
 
+First install all node dependencies:
+
+```
+npm install
```
nonowazu, line 13: Maybe I'm being pedantic or elitist here – any love for `yarn` here?

README.md
```diff
@@ -2,14 +2,31 @@
...
 
 ```
-hugo server -s exampleSite/ --themesDir=../.. --disableFastRender
+npm run serve:hugo
 ```
 
 ## Using the editor locally
 
-cd to the `exampleSite` directory and run `npx netlify-cms-proxy-server` alongside the `hugo` command; afterwards, navigate to `http://localhost:1313/admin`
+Change `IS_LOCAL_BUILD` inside `cms/index.js` to `true`.  This makes sure, netlify cms is available for your local build.
```
nonowazu, line 24: I know I'm nitpicking the docs, but this is important – what about setting an environment variable for this? Then you could do something like `yarn build-cms` or `yarn build-cmd:dev` (or the appropriate `npm run` substitution).

assets/css/main.css
```diff
@@ -79,26 +79,30 @@ Content
...
 
 .role-header,
-.role-header-healer,
-.role-header-tank,
-.role-header-dps {
+.role-header-healers,
+.role-header-tanks,
+.role-header-melee,
+.role-header-ranged,
+.role-header-casters {
```
nonowazu, line 86: I think this is where I pull out the 🔪 and point out that these changes are separate from generating editor config

exampleSite/content/jobs/healers/_index.md
```diff
@@ -3,6 +3,7 @@ menu:
   main:
     name: Healers
     parent: jobs
-role: healer
+role: healers
+resources_background_image: /img/healers_ressources.png
```
nonowazu, line 7: There's that typo for `resources` again :-)

nonowazu (changes requested):
I need a bit to review the javascript a bit more, but I did have some nits.

OHaiDanny:
@Roxas92 I'm doing some QA testing on the feature. One thing I noticed is in the BIS CMS section when I added in a new set, it doesn't show up on the web page.

image

image

OHaiDanny:
Ran into a Hugo server error when I tried to add content into the "Job Changes" section.

image

The error:
image


Something to do with the Time variable?

nonowazu:
More feedback based on local testing:

1. Since we know what job each landing pages if for, should we pre-fill out this string, and mark it as hidden?
Screen Shot 2021-11-10 at 6 56 12 PM

2. I'm debating if this should be optional or not. Please discuss?
Screen Shot 2021-11-10 at 6 57 41 PM

3. I feel like this is going to get really full really quick and `bg.png` might not cut it. Maybe we should make separate folders for each job?
Screen Shot 2021-11-10 at 6 58 25 PM

nonowazu:
> Ran into a Hugo server error when I tried to add content into the "Job Changes" section.
> 
> image
> 
> The error: image
> 
> Something to do with the Time variable?

I believe this is because that's missing the `lastmod` field, which the `changes.html` layout/template [uses](https://github.com/The-Balance-FFXIV/glam/blob/dd1451338d65e03d672f6d43cf4180ab88e2817d/layouts/jobs/changes.html#L11)

assets/css/main.css
Roxas92: I can understand your concerns, but please also keep in mind that the admin cp changes finally integrated the roles correctly. If I would not integrate this change into the PR, it would mean that we have visually speaking a more "broken" state that we have without this PR. So the 🔪 might be a bit harsh here. Take a ⚾ or so instead.

nonowazu: I think my concerns lie in a few places:

1. I'm not sure we're going to land these editor changes in production before soft launch
2. The editor and the theme are two separate components in terms of purpose. Absolutely – one does influence the other, but the theme can absolutely work without the editor.
3. Combined PRs can hold up changes that we need because we need to wait for all part(s) to be ready for merging.

Roxas92:
I will close this PR because there are too many conflicts been caused by the reason I made too many changes to files which were not affecting the admin config refactoring itself. I will think about that again and maybe opening a new one.

PR #90, opened 2021-11-11
Remove SMN for soft launch per request

OHaiDanny:
As per @lailleo request, this PR removes SMN in the role slider. We'll add this back once the content is done for future launch.


nonowazu:
This won't affect the production content once we move this role-slider config editor config into `balance-static` – I think we should close this PR and add the role slider editor config into the content repo and fill that out (sans smn)

OHaiDanny:
Ok sounds good. I'll close this and make the pr edit in the `balance-static`

Issue #87, opened 2021-11-10
editor example(s)


nonowazu:
- [ ] Role landing page
  - [ ] Should either hard-code or have a drop-down for the role
  - [ ] Body text might be nice, but not required
- [ ] Job landing page
  - [ ] Should probably hardcode the title (or set a param and change templates) since icons are dependent upon it

OHaiDanny:
I'll take a stab at this one.

OHaiDanny:
@nonowazu for the job landing page are you referring to the big "Jobs" title?

![image](https://user-images.githubusercontent.com/6521857/141405661-afa27ae8-6942-4b12-bab7-735af73d1bf0.png)

PR #92, opened 2021-11-12
Editor Changes for Role landing pages for issue #87

OHaiDanny:
@nonowazu in this PR i've added the editor changes for roles. I hope I understood the task correctly.. #87 

Couple clarifying things:

1. The names of the roles should be in plural form?
2. the role for melee, ranged, caster dps should be identifed as `dps`?

PR #93, opened 2021-11-12
Editor changes for Job landing page for issue #87

OHaiDanny:
@nonowazu probably need some feedback on this one. I think this is what you are looking for. #87 

After setting this up i'm thinking that there will probably be another landing page for non-combat jobs (DoH and DoL) in the future. Right now this is currently set to add combat jobs but it is something to think about maybe later.


exampleSite/static/admin/config.yml
```diff
@@ -30,6 +30,16 @@ collections:
           - {label: "Name", name: "name", widget: "string"}
           - {label: "Link", name: "job_link", widget: "string"}
           - {label: "Icon", name: "icon", widget: "string"}
+  # job landing page
```
nonowazu, line 33: Assuming this is linked to #87:

1. The "Role Landing Page" bullets refer to `/content/jobs/${role}/_index.md` – for now, we only need to have a selection for the role (hard-coded since this is going to end up being a file collection just like other job example(s))
2. If you *wanted* to add body text, I wouldn't stop it.

If this is referring specifically to `/content/jobs/${role}/${job}` bullets, then we need to hardcode the title (or the `job_name` value that Scrubby is going with.

nonowazu (changes requested):

exampleSite/static/admin/config.yml
OHaiDanny: Hey @nonowazu 

1. I have the changes in the other PR here: https://github.com/The-Balance-FFXIV/glam/pull/92. I believe it addresses your comments
2. I'll take a look at this again after reviewing Scrubby's changes.

OHaiDanny:
@nonowazu I've updated the PR. Here is what the CMS looks like: 

image


Inside each are three fields for `name` and `job_name`. The `body` is an optional field.

OHaiDanny:
Also to note is I used the CMS to generate all the `_index.md` files

exampleSite/static/admin/config.yml
OHaiDanny: @nonowazu i've updated this PR.

nonowazu (approved):

Issue #94, opened 2021-11-12
Changing asset locations breaks the role slider


nonowazu:
Should be a quick in the data files:
<img width="1627" alt="Screen Shot 2021-11-12 at 9 42 31 AM" src="https://user-images.githubusercontent.com/83298827/141493786-48f66a31-dd4e-4664-b08f-41965895b866.png">

ScrubbyXIV:
Ye, should be. That's something I only realised this morning and should've mentioned in the other PR. I'll work on the fix now.

nonowazu:
Fixed by #96

PR #96, opened 2021-11-12
Asset location bugfix

ScrubbyXIV:
- fix stuff for #94 that I broke
  - fix the icon location for all jobs
  - fix the job header because the change broke for all job guide pages


ScrubbyXIV:
@nonowazu if you have time, please verify if the header fix solved an issue I introduced yesterday for all job guides (fundamentals, etc.), thanks <3

PR #99, opened 2021-11-12
Comment out role resources in job pages

### LINKED ISSUE #97: Remove "job resources" from job landing page template
Since it's hardcoded and not functional, it's probably weird for a public launch:
Screen Shot 2021-11-12 at 12 50 03 PM
>

ScrubbyXIV:
Resolves #97 
- hide role resources because nono said no
- Removed it from other pages as well, as it wouldn't make sense to keep them anywhere but the job landing page


nonowazu (approved):
easy peazy lemon squeegee or what is squeezy why am I so wheezy?

PR #104, opened 2021-11-12
Remove redundant lock files and refined starting docs.

ArcaneDisgea:
Gutted npm in lieu of yarn.

I also tweaked the docs to add some missing information. Yarn.lock also freaked out because windows and different line endings I guess.


nonowazu (approved):
Thank you for looking through this!

Issue #108, opened 2021-11-14
Param for copyright statement


nonowazu:
It's technically bad to launch without a copyright statement, but maybe we can slip it in as a param (and leave blank otherwise) so it's a smooth transition.

PR #120, opened 2021-11-14
Copyright message

OHaiDanny:
Issue #108. Let me know if this is the right spot for the params. I'm wondering if this is really needed to be in the `balance-static` repo. Will it read the params once everything is merged?


layouts/partials/footer.html
```diff
@@ -33,7 +33,7 @@ <h3 class="footer-header pb-1">
       {{ end }}
     </div>
     <div class="py-5 footer-copyright">
```
nonowazu, line 35: Maybe put a

```html
{{ with .Site.Params.copyrightMessage }}
<div class="py-5 footer-copyright">
<!-- the rest -->
{{ end }}
```

nonowazu (changes requested):
Just oooone small change

layouts/partials/footer.html
OHaiDanny: will do!

OHaiDanny: with broke the site. but If statement works :)

OHaiDanny:
I also fixed some of the footer margin spacing. just some small CSS changes

layouts/partials/footer.html
nonowazu: Interesting, but quick fix if we get it to work with `with` later.

PR #129, opened 2021-11-15
Rename black mage background

OHaiDanny:


lailleo (approved):
Looks correct per discussion in dev channel.

Issue #113, opened 2021-11-14
table of contents only looks at h2 and below?


nonowazu:
Whyyyyyyyyy?

OHaiDanny:
Hugo Table of Contents doesn't support h1 :(

https://gohugo.io/content-management/toc/

lailleo:
#133

PR #133, opened 2021-11-15
Add ToC config and styling for nested sections

hintxiv:
Addresses #113 

I still don't know how to run the CMS locally but here's some sample markdown: 

image

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -68,6 +68,15 @@ Heading
 
 
 /**
+Table of Contents
+ */
+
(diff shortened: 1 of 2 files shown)


OHaiDanny:
@hintxiv to run the local CMS you simply run `yarn start:admin` (i recommend opening up a new terminal tab and do this as you have your localhost running). If successful you should see this:

image

Then you can go to your localhost /admin and it should log you into the CMS.

nonowazu (approved):

Issue #131, opened 2021-11-15
Pages need titles to display in browser tabs


lailleo:
Beautification.

OHaiDanny:
PR created.

nonowazu:
Addressed by #134

PR #134, opened 2021-11-15
Add <title> tag in head

OHaiDanny:
@lailleo i think you mentioned in one of the issues that the <title> tag is missing for browser title? this should fix it.

Fix for issue #131

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -8,6 +8,7 @@
 <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#ff9900">
 <meta name="msapplication-TileColor" content="#ffffff">
 <meta name="theme-color" content="#ffffff">
+<title>{{ if .Title }}{{ .Title }} | {{ end }}{{ .Site.Title }}</title>
 <link rel="preconnect" href="https://fonts.googleapis.com">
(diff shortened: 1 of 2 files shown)


layouts/partials/head.html
```diff
@@ -1,6 +1,7 @@
 <meta http-equiv="Content-Type" content="text/html" charset="UTF-8" />
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
 <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+<title>{{ if .Title }}{{ .Title }} | {{ end }}{{ .Site.Title }}</title>
```
nonowazu, line 4: Can we do `{{ with .Title }}` here?

Also, Maybe make the title something like:

```
{{ .Site.Title }} {{ with .Title }} | {{ . }} {{ end }}
```

That way the tab would be `The Balance | Monk BiS`. Unless we should have the page title first before the `|`?

nonowazu (changes requested):

layouts/partials/head.html
OHaiDanny: cool cool!

nonowazu (approved):

nonowazu:
Oh wait, there's some head partial conflicts you need to resolve.

PR #138, opened 2021-11-15
fix header name issue with underscore

ScrubbyXIV:
Fixes header job names for subpages of jobs.

### DIFF
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -6,7 +6,8 @@
   {{ $header := dict
     "role" $role
     "job" .Parent.Params.job_name
-    "title" "Best-in-slot (BIS) Gear Sets"
+    "jobTitle" .Parent.Title
+    "title" .Params.title
     "icon" "/theme-assets/armoury.svg"
     "lastmod" .Params.lastmod
     "patch" .Params.patch
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -7,7 +7,7 @@
     "role" $role
     "job" .Parent.Params.job_name
     "jobTitle" .Parent.Title
-    "title" .Params.title
+    "title" "Best-in-slot (BiS) Gear Sets"
     "icon" "/theme-assets/armoury.svg"
     "lastmod" .Params.lastmod
     "patch" .Params.patch
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -7,7 +7,7 @@
     "role" $role
     "job" .Parent.Params.job_name
     "jobTitle" .Parent.Title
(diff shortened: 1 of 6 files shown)


layouts/partials/job/single_header.html
```diff
@@ -1,11 +1,11 @@
-{{ $jobName := (lower (replace .job " " "_")) }}
-<div class="h-60 xl:h-80 bg-no-repeat bg-cover" style='background-image: url(/theme-assets/jobs/{{ lower .role }}/{{ $jobName }}/{{ $jobName }}_background.png);'>
+{{ $jobName := (lower (replace .job "_" " ")) }}
```
nonowazu, line 1: Wouldn't it be better to get the jobname from [here](https://github.com/The-Balance-FFXIV/glam/blob/ac77d21174ab7ca4a382402f2d8032b46b017a62/layouts/jobs/bis.html#L8)? Or even doing `.Parent.Title`?

nonowazu (changes requested):

layouts/partials/job/single_header.html
ScrubbyXIV: job (taken from job_name) contains the underscore. The .Parent.Title one could work though, yes. Should I change it to that?

ScrubbyXIV: Actually, it doesn't seem to work locally...

nonowazu: I would add another parameter to the header dict, and just use that – it's more or less guaranteed to be correct in the `balance-static` editor.

nonowazu: Of course, this also means updating all the templates, but I don't think that's awful all things considered.

ScrubbyXIV: it's only a couple of files, ye. I'll go with that then.

ScrubbyXIV: Adding the parameter and using the .Parent.Title should not introduce any breaking changes, so we should be fine.

layouts/jobs/bis.html
```diff
@@ -6,7 +6,8 @@
   {{ $header := dict
     "role" $role
     "job" .Parent.Params.job_name
-    "title" "Best-in-slot (BIS) Gear Sets"
+    "jobTitle" .Parent.Title
+    "title" "Best-in-slot (BiS) Gear Sets"
```
nonowazu, line 10: One final nit – this can probably be `.Title`, since we have a title prop on all pages now (I should know, I did the [PR for it](https://github.com/The-Balance-FFXIV/balance-static/pull/725))

nonowazu (changes requested):

nonowazu (approved):

PR #139, opened 2021-11-15
BiS title change

ScrubbyXIV:
change BiS title from hardcoded to parameter

### DIFF
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -6,7 +6,7 @@
   {{ $header := dict
     "role" $role
     "job" .Parent.Params.job_name
-    "title" "Best-in-slot (BIS) Gear Sets"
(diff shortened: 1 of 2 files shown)


layouts/jobs/bis.html
```diff
@@ -6,7 +6,7 @@
   {{ $header := dict
     "role" $role
     "job" .Parent.Params.job_name
-    "title" "Best-in-slot (BIS) Gear Sets"
+    "title" .Params.title
```
nonowazu, line 9: I think this should just be `.Title` instead of `.Params.title` – this is one of the default frontmatter items in hugo, and becomes a proper variable without needing `.Params`

nonowazu (changes requested):
Small change requested

nonowazu (approved):

Issue #111, opened 2021-11-14
Make sure H1 markdown matches the mockups


lailleo:
Was added quickly to fix for soft launch, but will need to be beautified to match Udra and Moth's mockups.

OHaiDanny:
I think we can't use h1 since there are some hugo features (like table of conents) where it does not detect a h1 tag.

lailleo:
We might need to make H2 match it, in that case.

OHaiDanny:
With the recent discovery by @hintxiv, we can now use H1 with Table of Contents. I'll update the styles for it to match the mockup from frontend pins 😄

PR #149, opened 2021-11-16
H1 styling

OHaiDanny:
This address issue #111


OHaiDanny:
I recommend merge Hint's #150 PR first then mine. I think the only difference i

PR #156, opened 2021-11-16
ToC Styling

hintxiv:
- Move the table of contents to its own partial
- Don't render ToC if it's empty
- Add smooth scrolling when you click on a link
- Add current section highlighting in white (per #frontend) 
- Reduce font size (per #frontend) 

https://streamable.com/af2gbk

### DIFF
--- a/assets/js/tableOfContents.js
+++ b/assets/js/tableOfContents.js
@@ -1,0 +1,43 @@
+function tocItem(anchor) {
+  return document.querySelector("[href=\"" + anchor + "\"]")
+}
+
+function heading(anchor) {
+  return document.querySelector("[id=" + anchor.substr(1) + "]")
+}
+
+var _anchors = null
+function anchors() {
+  if (!_anchors) {
+    _anchors = Array.from(document.querySelectorAll("#TableOfContents a"))
+        .map(e => e.getAttribute('href'))
+  }
+  return _anchors
+}
+
+function currentAnchor() {
+  var bestAnchor = null
+  var bestY = null
+  anchors().forEach(anchor => {
+    var y = heading(anchor).getBoundingClientRect().top
+    if (y >= 0 && (!bestAnchor || y < bestY)) {
+      bestAnchor = anchor
+      bestY = y
+    }
+  })
+  return tocItem(bestAnchor)
+}
+
+function update() {
+  var selected = document.getElementById("tocSelected")
+  var anchor = currentAnchor()
+  if (selected) {
+    selected.removeAttribute('id')
+  }
+  if (anchor) {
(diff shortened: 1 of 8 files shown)


layouts/jobs/single.html
```diff
@@ -23,9 +23,7 @@
         </div>
       </div>
       <div class="col-span-3 bg-card-light text-white divide-y divide-page px-5 sticky top-8">
-        <div class="card-title pt-6 pb-5">Table of Contents</div>
-        <div class="nested-link font-bold py-2">{{ .TableOfContents }}</div>
-        <div>&nbsp;</div>
+        {{ partial "components/toc.html" . }}
```
nonowazu, line 26: We might need to revisit this since we might have alternative toc partial(s) for the changes, faq, and bis templates.

hintxiv: The FAQ ToC in particular I have no idea how we should handle.

For the rest of them, is there a conventional way to map different partials to their corresponding layout? I'm fine with breaking just it out into separate files (maybe `partials/components/toc/changes_toc.html`, `.../bis_toc.html`, etc.) if the need arises.

nonowazu: I think we'll just need perhaps a `components/toc.html` and a `components/custom_toc.htm`? I think the onus of formatting stuff into a slice/dict/whatever should be on the layouts/partials that call them.

assets/js/tableOfContents.js
```diff
@@ -0,0 +1,43 @@
...
+  if (selected) {
+    selected.removeAttribute('id')
+  }
+  if (anchor) {
+    anchor.setAttribute('id', 'tocSelected')
+  }
+}
+
+window.onload = update
+window.onscroll = update
```
nonowazu, line 43: I think we need to look more at the [intersection observer api](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) for this, since it's far more performant and doesn't involve globally editing window properties in a script.

nonowazu (changes requested):

layouts/partials/components/qna_toc.html
```diff
@@ -0,0 +1,17 @@
...
+    <ul>
+      {{ range $index, $q := . }}
+      <li>
+        <a href="#q{{ $index }}">{{ $q.question }}</a>
+      </li>
+      {{ end }}
+    </ul>
+  </nav>
+</div>
+<div>&nbsp;</div>
```
hintxiv, line 17: @nonowazu this works fine for the qna template specifically. I honestly have no idea how to go about dynamically building a dict of headers with Hugo in order to generalize this partial beyond qna.

nonowazu (approved):

PR #157, opened 2021-11-16
add OpenGraph and regular Meta

ArcaneDisgea:
I have no real way of testing but as far as I can tell this should Just Work™️


ArcaneDisgea:
Yo I was able to test this and repush my changes. So two things to note is we want to take advantage of the keywords meta for SEO purposes I guess idk and that will be just be a list of tags. This can be done per page it goes write in the guides md file.

```
tags:
  - scholar guide
  - ffxiv scholar
  - endwalker guide
  - foo
  - bar
  - baz
```

Next is the image for the embed that is done using a list of images that it may or may not pick randomly from? I wasn't able to find a clear answer on this nor could I get it to choose any image other than the first so maybe the others are second fallback images in case first fails not sure.

```
images:
  - /theme-assets/jobs/healers/scholar/scholar.png
```

The template image I chose isn't necessarily ideal but its just a placeholder to prove it works.

image

nonowazu:
I like this a lot, but we might want to edit the example editor to have the description fields and tags.

ArcaneDisgea:
The description fields are already pulled from the md I believe unless I am misunderstanding.

nonowazu:
Sorry for the slow response, I mean add `description` fields to `exampleSite/static/admin/config.yml`.

ArcaneDisgea:
All good, should I do that for the image/tags as well? I assume it's just adding a line like this...

```
baseURL = "https://example.com"
languageCode = "en-us"
defaultContentLanguage = "en"
title = "The Balance"
theme = "glam"
description = "Glam might actually be the best hugo theme there is."
```

nonowazu (approved):

PR #171, opened 2021-11-18
Job Guides Pages (Single, BiS, QnA, Job Changes) mobile responsiveness

OHaiDanny:
All these pages are good to go for a first pass.

<strike>The only thing I need to check are some various mobile Table tag styling.</strike>

Table styling will be in a new issue (it's going to get tricky styling each table we have)

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -18,7 +18,7 @@ body {
 *  Container (Mobile + Desktop)
 */
 .responsive-container {
-	@apply container px-4 lg:px-8 xl:px-0 mx-auto;
+	@apply container px-4 lg:px-8 xl:px-4 mx-auto;
 }
 
 /*
@@ -200,7 +200,7 @@ Content
 }
 
 .meta-content {
-	@apply hidden bg-card-light px-4 py-6 w-full;
+	@apply hidden bg-card-dark px-4 py-6 w-full;
 }
 
 /*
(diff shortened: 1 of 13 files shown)


assets/css/main.css
```diff
@@ -153,31 +153,22 @@ Content
...
-	margin-bottom: -1px;
-	font-size: 1.5rem; 
-	font-weight: 600;
-	text-transform: uppercase;
-	letter-spacing: 1px;
-	transition: all 0.3s;
-	background: #1E1E1F;
-	border-right: 2px solid #141415;
-}
-
```
OHaiDanny, line 174: There's duplicate styling below

nonowazu: Thanks for the explanation ahead of time by the way.

layouts/partials/components/questions.html
```diff
@@ -0,0 +1,10 @@
+<div class="card-title mb-5">Have any questions?</div>
+  <a href="https://discord.gg/thebalanceffxiv">
+    <div class="flex flex-col md:flex-row items-left md:items-center mb-9">
+      <img class="mr-4 mb-2" src="/theme-assets/the_balance.png">
+      <div class="flex items-center">
+        <img class="mr-2" src="/theme-assets/orange_discord_no_text.png">
+        <span class="text-link-orange">discord.gg/thebalanceffxiv</span>
```
nonowazu, line 7: aaaah, we need a discord param so we can use that in places like this

assets/css/main.css
```diff
@@ -219,6 +226,12 @@ It unfortunately also means we're limited to 5 tabs at current.
 	@apply block bg-card-light flex content-center xl:h-120;
 }
 
+.tabbed [type="radio"]:nth-of-type(1):checked ~ .tabs > .tab:nth-of-type(1) > .meta-content,
+.tabbed [type="radio"]:nth-of-type(2):checked ~ .tabs > .tab:nth-of-type(2) > .meta-content,
+.tabbed [type="radio"]:nth-of-type(3):checked ~ .tabs > .tab:nth-of-type(3) > .meta-content {
```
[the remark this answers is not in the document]
OHaiDanny: I'll revisit this again. I was trying to think how best to reuse it. The main challenge is that the role slider content has a fixed height and if I try to reuse it it will make these meta-content stretch out waaay to long.

layouts/partials/components/questions.html
OHaiDanny: for now, can it be hard-coded? or the discord link pay change 😮

OHaiDanny: oh you mean a param in like the config.toml or something

OHaiDanny: ill also have to replace the href in the a tags with this new param too maybe?

nonowazu (approved):

PR #176, opened 2021-11-19
[Mobile] Responsive container on xl screens missing left right padding

OHaiDanny:
This fixes Bongo's Kitten no. 69's issue regarding 1280px and above missing left right spacing.

### DIFF
--- a/layouts/partials/header_home.html
+++ b/layouts/partials/header_home.html
@@ -1,7 +1,7 @@
 {{ $current := . }}
 {{ $list := .Site.Menus.main | append .Site.Menus.header }}
-  <nav class="fixed lg:absolute top-0 left-0 w-full z-50">
-    <div class="container mx-auto">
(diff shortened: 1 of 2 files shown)


layouts/partials/header_home.html
```diff
@@ -1,7 +1,7 @@
 {{ $current := . }}
 {{ $list := .Site.Menus.main | append .Site.Menus.header }}
-  <nav class="fixed lg:absolute top-0 left-0 w-full z-50">
-    <div class="container mx-auto">
+  <nav class="fixed lg:absolute top-0 left-0 w-full z-50 bg-card-light lg:bg-transparent">
```
OHaiDanny, line 3: While testing i noticed this weird effect on the homepage mobile menu. This should fix it.

nonowazu (approved):

PR #180, opened 2021-11-19
Mobile landscape size adjustments across the board.

OHaiDanny:
This PR fixes various size issues that is giving landscape mode a bad experience.

### DIFF
--- a/layouts/partials/header_home.html
+++ b/layouts/partials/header_home.html
@@ -1,7 +1,7 @@
 {{ $current := . }}
 {{ $list := .Site.Menus.main | append .Site.Menus.header }}
-  <nav class="fixed lg:absolute top-0 left-0 w-full z-50">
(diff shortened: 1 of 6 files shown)


layouts/partials/header_home.html
```diff
@@ -1,7 +1,7 @@
 {{ $current := . }}
 {{ $list := .Site.Menus.main | append .Site.Menus.header }}
-  <nav class="fixed lg:absolute top-0 left-0 w-full z-50">
-    <div class="container mx-auto">
+  <nav class="fixed lg:absolute top-0 left-0 w-full z-50 bg-card-light lg:bg-transparent">
+    <div class="responsive-container">
```
OHaiDanny, line 4: these two lines are the same changes from #176. I needed to add it here for me to test

nonowazu (approved):

PR #152, opened 2021-11-16
Add <head> tags to file to rendered pages

kaydeethree:
Rendered pages are not generating valid HTML. They also do not have <title> tags but this will take more effort to resolve

PR #183, opened 2021-11-19
add head tags to baseof

nonowazu:
I left feedback in PR #152, but I didn't hear back. So I made this one instead.


ArcaneDisgea (approved):
There are no head tags inside the partial itself so it looks good to me. 

I'm not sure what Hugo's style standards are but I personally typically just include the head tag inside the partial itself that way I can just include the partial and done.

PR #187, opened 2021-11-20
Menu rendering fixes

nonowazu:
This PR:

• Moves non-main-config.toml specific menu config into the appropriate frontmatter
• Changes the hover menu items to actually use the URL.

Also, as a note – the mobile menu is only head in `header_home.html` and not just in regular header – we should partial-itize it and make it available on both partials.


Lunaversitay:
> This PR:
> 
> • Moves non-main-config.toml specific menu config into the appropriate frontmatter • Changes the hover menu items to actually use the URL.
> 
> Also, as a note – the mobile menu is only head in `header_home.html` and not just in regular header – we should partial-itize it and make it available on both partials.

Yeah will definitely make it into a partial and add it to the second navbar - whenever you're able to push this I'll pull it to my env and work on that 👍

PR #195, opened 2021-11-20
Bugfix on Stat priority no content

OHaiDanny:
This PR fixes the issue where we hide the link to the stat priority page if there is no content.

image

### DIFF
--- a/layouts/partials/cards/default.html
+++ b/layouts/partials/cards/default.html
@@ -15,8 +15,10 @@
     <div>
       {{ .text }}
     </div>
-    {{ if isset . "readMoreLink" }}
-      {{ partial "cards/_footer.html" . }}
-    {{ end }}
+    <div class="border-t border-card-dark pt-5 text-right">
+      {{ if isset . "readMoreLink" }}
+        {{ partial "cards/_footer.html" . }}
+      {{ end }}
+    </div>
 	</div>
 </div>
--- a/layouts/partials/job/gearing.html
+++ b/layouts/partials/job/gearing.html
@@ -21,25 +21,25 @@
       </div>
       <div class="xl:col-span-4 text-white">
         {{ with .Page.GetPage "stat-priority" }}
-          {{ if .Content }}
-            {{ $content := dict
+          {{if .Content }}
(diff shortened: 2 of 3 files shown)


layouts/partials/job/gearing.html
```diff
@@ -21,15 +21,25 @@
...
+            {{ partial "cards/default.html" $content }}
+          {{ else }}
+            {{ $content := dict
+              "role" $role
+              "icon" "/theme-assets/materia.svg"
+              "name" "Stats & Materia"
+              "text" .Params.priority
+            }}
+            {{ partial "cards/default.html" $content }}
+          {{ end }}
```
OHaiDanny, line 42: I kinda dont like how im repeating code. I was trying to find out if Hugo has a way to set something like a "Global Variable" so I just have to update that in the IF statement...

layouts/partials/cards/default.html
```diff
@@ -15,8 +15,10 @@
     <div>
       {{ .text }}
     </div>
-    {{ if isset . "readMoreLink" }}
-      {{ partial "cards/_footer.html" . }}
-    {{ end }}
+    <div class="border-t border-card-dark pt-5 text-right">
+      {{ if isset . "readMoreLink" }}
```
OHaiDanny, line 19: Moving the container outside this file will create an empty space when we don't render a link in the `dict`

layouts/partials/job/gearing.html
nonowazu: I think you should do a parent dict and then call "merge" on the extra options perhaps? Docs [here](https://gohugo.io/functions/merge/)

nonowazu (changes requested):

layouts/partials/job/gearing.html
OHaiDanny: I found another solution to use `.Scratch` instead to solve it. the `merge` has a problem of scoping when it is inside if statements unfortunately...

https://www.regisphilibert.com/blog/2017/04/hugo-scratch-explained-variable/

nonowazu (approved):

PR #199, opened 2021-11-21
feature/footer-hide-nav

Hughes89:
-  If a footer menu item doesn't contain any children (links) then don't render it.


zorianak:
Looks like this is now pretty out of date/old. In addition, the footer item that had no children has been fixed now.

Issue #200, opened 2021-11-22
Implement Etro embed


lailleo:
Documentation: 

https://etro.gg/docs/embed

Nono also asked me to mention:

"mention that we need a partial for it, and that having a shortcode wouldn't be bad either."

ductapeofdoom:
#202 addresses this.

rexatlanti:
inplemented in #202

PR #202, opened 2021-11-22
Added a partial and shortcode to support etro.gg embeds

ductapeofdoom:
Addresses #200.

Also included a commented out example in the `bis.html` partial for later use if needed.

I was unsure of the exact size desired so feedback on the width/height of the embed is greatly appreciated. For reference, the current height is 96 rem.

### DIFF
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -21,8 +21,6 @@
         <div class="markdown max-w-none">
           {{ range .Params.bis }}
             <h2><a href="{{ .link }}">{{ .name }}</a></h2>
-            <!-- {{ $link := dict "uuid" (index (last 1 (split .link "/")) 0) }} 
-            {{ partial "components/etro.html" $link}} -->
             {{ .description | markdownify }}
           {{ end }}
         </div>
--- a/layouts/jobs/bis.html
+++ b/layouts/jobs/bis.html
@@ -21,6 +21,8 @@
         <div class="markdown max-w-none">
           {{ range .Params.bis }}
             <h2><a href="{{ .link }}">{{ .name }}</a></h2>
+            <!-- {{ $link := dict "uuid" (index (last 1 (split .link "/")) 0) }} 
+            {{ partial "components/etro.html" $link}} -->
             {{ .description | markdownify }}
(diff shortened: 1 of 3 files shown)


ductapeofdoom:
https://i.imgur.com/A9rGhrn.png example image

layouts/jobs/bis.html
```diff
@@ -21,6 +21,8 @@
         <div class="markdown max-w-none">
           {{ range .Params.bis }}
             <h2><a href="{{ .link }}">{{ .name }}</a></h2>
+            <!-- {{ $link := dict "uuid" (index (last 1 (split .link "/")) 0) }} 
+            {{ partial "components/etro.html" $link}} -->
```
nonowazu, line 25: While I appreciate this, I think we should remove this from there since `<!-- -->` doesn't stop it from calling the partial(s).

We should probably also make a `/docs` directory and put a `shortcodes.md` to explain all the shortcode(s) we have, and how to use them.

layouts/partials/components/etro.html
```diff
@@ -0,0 +1,6 @@
+<div class="h-96">
+  <iframe class="w-full h-full"
+    title="Etro Gearset" 
+    src="https://etro.gg/embed/gearset/{{ .uuid }}">
```
nonowazu, line 4: since we're only dealing with one value here, instead of creating a dictionary just so we can have a `.uuid` field, maybe it's appropriate to *only* pass in `$uuid` and use it here as just `.`?

nonowazu (changes requested):
Looks good, just a little bit of frontloaded work since you're the first to land a shortcode here — sorry!

layouts/jobs/bis.html
ductapeofdoom: Good call! Will delete the comment.

layouts/partials/components/etro.html
ductapeofdoom: Yeah I originally had more parameters being passed, but removed them and forgot to remove the dict.

ductapeofdoom:
Updated PR based on feeback.

nonowazu (approved):

PR #213, opened 2021-11-23
Fix missing `">`

nonowazu:
Title ☝️


layouts/partials/head.html
```diff
@@ -5,7 +5,6 @@
 <link rel="shortcut icon" sizes="192x192" href="/android-chrome-192x192.png">
 <link rel="shortcut icon" sizes="512x512" href="/android-chrome-512x512.png">
 <link rel="apple-touch-icon" href="/apple-touch-icon.png">
-<meta name="viewport" content="width=device-width, initial-scale=1.0
```
nonowazu, line 8: This line is available higher up in the file.

PR #216, opened 2021-11-24
Mobile Navigation

OHaiDanny:
This PR adds mobile and tablet navigation based on the mockups.

This features a chevron down icon next to the parent navigation. Clicking on the chevron will expand to show sub-menus. Clicking on the name of the parent will direct you to that page.

image


assets/js/nav.js
```diff
@@ -0,0 +1,17 @@
+const navChevron = document.querySelectorAll(".chevron-down");
+const chevronUp = "/theme-assets/chevron-up.svg";
+const chevronDown = "/theme-assets/chevron-down.svg";
+
+for (let i = 0; i < navChevron.length; i++) {
```
nonowazu, line 5: since the index itself isn't required ('cause we're not doing any `foo[i+1]`-style neighbor searching), I would simplify this to

```js
for (const element of document.querySelectorAll('...')) {
```
Since (imo):

* It saves an extra const declaration that won't be used later anyway
* makes it clear that you're not going to replace any of the elements you find
* makes the code a bit more readable than `navChevron[i]` (you could replace `element` with `navElement` if that makes it more readable though)

assets/js/nav.js
```diff
@@ -0,0 +1,17 @@
...
+    let siblingList = navChevron[i].nextElementSibling;
+    navChevron[i].addEventListener('click', () => {
+        if (navChevron[i].children[0].src.indexOf('chevron-down') >= 0) {
+            navChevron[i].children[0].src = chevronUp
+        } else { 
+            navChevron[i].children[0].src = chevronDown
+        }
+        navChevron[i].parentNode.classList.toggle('nav-active');
+        siblingList.classList.toggle('hidden')
+    });
```
nonowazu, line 15: I feel like a lot of this code is brittle; why not use `navElement` (as discussed earlier) as `navElement.querySelector`? It's already scoped to the parent element and lets you search down.

If we restructure the html a little bit, this could be simplified to toggle on/off an `.active` class that would just change the chevron (and turn on/off hiding), which might be a tad easier on your sanity.

layouts/partials/head.html
```diff
@@ -20,7 +20,9 @@
 <link href="https://fonts.googleapis.com/css2?family=Kumbh+Sans:wght@700&family=Roboto:ital,wght@0,400;0,700;1,400&display=swap" rel="stylesheet">
 {{ $styles := resources.Get "css/main.css" | postCSS }}
 {{ $headerJs := resources.Get "js/header.js" }}
+{{ $navJs := resources.Get "js/nav.js" }}
 <script type="text/javascript" src="{{ $headerJs.RelPermalink }}" defer></script>
+<script type="text/javascript" src="{{ $navJs.RelPermalink }}" defer></script>
```
nonowazu, line 25: This is not a 'now' thing, but a "Someone please remind me later" thing — we should move these to the bottom of the page so that they're not blocking on load. We'll need a partial called "extraScripts" or something. Either way, someone please remind me later when we're doing a polishing pass.

layouts/partials/header.html
```diff
@@ -42,13 +42,34 @@
...
             href="{{ .URL }}"
-            class="block text-sm px-6 py-2 md:py-3 text-white font-semibold {{ if eq $current.URL .URL }}text-link-orange{{ end }}" title="{{ .Title }}"
+            class="w-9/12 block text-xl px-6 py-2 md:py-3 text-white font-semibold {{ if eq $current.URL .URL }}text-link-orange{{ end }}" title="{{ .Title }}"
           >
             {{ .Name }}
           </a>
+          {{ if .HasChildren }}
+            <div class="chevron-down w-3/12 flex justify-end items-center text-right px-6 py-2 lg:hidden">
+              <img src="/theme-assets/chevron-down.svg" alt="expand list" />
+            </div>
```
nonowazu, line 55: I think if we embedded the `ul` in this `div`, added something like the styling I suggested above (forgoing the `img` and relying on css), moved the `hidden` to some `@apply` statement (and had a class that toggled the showing of such list), this would be a tad better. But let me know your thoughts?

nonowazu (changes requested):
All in all, this is good stuff. I just have some suggestions on minor changes to the html structure and css, and perhaps more drastic changes to the js.

layouts/partials/header.html
OHaiDanny: I was able to simplify this by using a div with background instead of the . It worked out fine :)

layouts/partials/head.html
OHaiDanny: I've created an issue for this

nonowazu (approved):

PR #236, opened 2021-11-29
Encounter Landing Page + Fight Guide Page

OHaiDanny:
This PR includes the addition of the encounter home page + fight guide page. See screenshots below.

Most of the content will be CMS driven. The CMS has been configured to support all the fields needed to generate the home page and the fight pages.

Home Page:
encounterHome

Fight Page
encounterSingle
image


exampleSite/static/admin/config.yml
```diff
@@ -307,23 +307,6 @@ collections:
         widget: "object"
         fields:
           - {label: "Discord Name", name: "discord_id", widget: "string"}
-  # encounter metadata
```
OHaiDanny, line 310: I don't think we need this anymore. We are not using anything from the `data` folder correct?

layouts/partials/encounter/homepage.html
```diff
@@ -0,0 +1,36 @@
...
+{{ $savage := where $encounterPages ".Params.encounter_category" "savage"}}
+{{ $ultimate := where $encounterPages ".Params.encounter_category" "ultimate"}}
+<div class="flex flex-col">
+  <div class="w-full mb-24">
+    <div class="flex flex-row items-end">
+      <h1 class="h-shadow flex-grow text-3xl sm:text-5xl uppercase tracking-wider font-head font-bold mb-8">Savage Tiers</h1>
+    </div>
+    <div class="w-full">
+      {{ range $savage.GroupByParam "tier_weight" }}
+        {{ range .Pages.GroupByParam "series_name"}}
```
OHaiDanny, line 11: Understanding how `PageGroups` and `Pages` work was a tough one to get this whole logic figured out

layouts/partials/encounter/homepage.html
```diff
@@ -0,0 +1,36 @@
...
+          <h2 class="text-2xl">{{ .Key }}</h2>
+        {{ end }}
+        <div class="grid grid-cols-4 gap-8 mb-8">
+          {{ range sort .Pages "Weight" "desc" }}
```
OHaiDanny, line 15: this sorts the weights in descending order. Each of the individual fights in the tier is sorted from weight 1-12. This sort scopes it so that I can add weights 1-12 on Omega tier fights, or Alexander Tier fights

layouts/partials/encounter/homepage.html
```diff
@@ -0,0 +1,36 @@
...
+            {{ partial "cards/encounters/fight_card.html" . }}
+          {{ end }}
+        </div>
+      {{ end }}
+    </div>
+  </div>
+  <div class="w-full mb-24">
+    <div class="flex flex-row items-end">
+      <h1 class="h-shadow flex-grow text-3xl sm:text-5xl uppercase tracking-wider font-head font-bold mb-8">Ultimates</h1>
```
OHaiDanny, line 24: I'll have to get rid of this. It shouldn't show a sub header when the main header is displaying `Ultimates`

layouts/partials/general_header.html
```diff
@@ -0,0 +1,3 @@
+<div class="general-page-bg bg-center bg-cover relative flex items-center text-center justify-center z-10 h-44 lg:h-96 mt-0">
```
OHaiDanny, line 1: this is some minor refactoring as I'm using the same header as the General Pages

layouts/partials/encounter/homepage.html
OHaiDanny: It is descending because fight 12 8 and 4 are on the left side of the screen.

exampleSite/content/encounters/uwu.md
```diff
@@ -0,0 +1,14 @@
+---
+title: The Weapon's Refrain
+card_image: /theme-assets/gaia-test.png
+tier_image: /theme-assets/gaia-test.png
```
nonowazu, line 4: Not a complaint on this or the path, but do we want to make this optional in the editor and use `default` in hugo to provide one from theme assets for before we get photos for a fight? This way we can get a day 1 guide up without having specific media assets.

exampleSite/content/encounters/uwu.md
```diff
@@ -0,0 +1,14 @@
...
+encounter_category: ultimate
+tier_name: Ultimate
+series_name: Ultimate
+weight: 2
+tier_weight: 1
+coming_soon: false
```
nonowazu, line 10: Do we want to also have a `spoiler` boolean so we could keep something marked as a spoiler for a few weeks after a guide launch? That way, `coming_soon` just adds the text, but `spoiler` enables the css blur. Just me wondering aloud at 2am.

exampleSite/static/admin/config.yml
nonowazu: Yeah. I suppose I had ideas about this before, but you are absolutely correct that data is probably too much overhead for something like this.

layouts/pages/single.html
```diff
@@ -1,10 +1,8 @@
 {{ define "main" }}
-    <div class="general-page-bg bg-center bg-cover relative flex items-center text-center justify-center z-10 h-64">
-        <h1 class="text-8xl h-shadow font-bold">{{ .Title }}</h1>
-    </div>
-    <div class="container mx-auto my-16 bg-card-light p-16">
-        <div class="markdown">
-            {{ .Content }}
-        </div>
-    </div>
+  {{ partial "general_header.html" . }}
```
nonowazu, line 2: nice savings on this addition!

layouts/partials/cards/encounters/fight_card.html
```diff
@@ -0,0 +1,17 @@
+<div class="relative mb-8">
+  {{ if eq .Params.coming_soon true}}
+  <a class="block bg-card-light">
+  {{ else }}
+  <a class="block bg-card-light" href="{{ .Permalink }}">
+  {{ end }}
+    <img class="block object-cover w-full" src="{{ .Params.tier_image}}" alt="" />
```
nonowazu, line 7: The smallest of nits: missing a space before the `}}`

layouts/partials/encounter/homepage.html
nonowazu: There is an actual ton of logic around page fetching, parsing, and metadata that, as I read through the docs, makes me think that the hugo people intentionally made it flexible like this so we could start treating pages as fetch-able DB resources

nonowazu (commented):
Honestly, I like what I see of everything so far, and while the page logic was harder to write, I applaud you for trucking through it and making it work so well (in surprisingly less templating than I thought — which I can also attribute to your previously good breakup of the base templates)

exampleSite/content/encounters/uwu.md
OHaiDanny: I can set up a `spoiler` boolean to create a blur effect. I think there is a CSS way to make it so.

exampleSite/static/admin/config.yml
```diff
@@ -307,23 +307,28 @@ collections:
...
-          - {label: "Image Link", name: "image_link", widget: "string"}
-          - {label: "Encounter Link", name: "encounter_link", widget: "string"}
+      - {label: "Title of Fight", name: "title", widget: "string"}
+      - {label: "Encounter Category", 
+        name: "encounter_category", 
+        widget: "select", 
+        options: ["savage", "ultimate", "extreme"]}
+      - {label: "Card Image", name: "card_image", widget: "image", choose_url: false}
+      - {label: "Tier Image", name: "tier_image", widget: "image", choose_url: false}
+      - {label: "Tier Name (Capitalized)", name: "tier_name", widget: "string"}
```
OHaiDanny, line 324: @nonowazu  Question on tier name field. Do you want me to list out all the values and put the in a dropdown so that we get the content team to manually type it in?

exampleSite/static/admin/config.yml
```diff
@@ -307,23 +307,28 @@ collections:
...
+      - {label: "Series Name (Capitalized)", name: "series_name", widget: "string"}
```
OHaiDanny, line 325: Same with this one.

exampleSite/static/admin/config.yml
```diff
@@ -307,23 +307,29 @@ collections:
...
-          - {label: "Encounter Link", name: "encounter_link", widget: "string"}
+      - {label: "Title of Fight", name: "title", widget: "string"}
+      - {label: "Encounter Category", 
+        name: "encounter_category", 
+        widget: "select", 
+        options: ["savage", "ultimate", "extreme"]}
+      - {label: "Fight Card Image", name: "card_image", widget: "image", choose_url: false}
+      - {label: "Fight Banner (for individual guides)", name: "banner_image", widget: "image", choose_url: false}
+      - {label: "Encounter Tier Image", name: "tier_image", widget: "image", choose_url: false}
+      - {label: "Encounter Tier Name (Capitalized)", name: "tier_name", widget: "string"}
```
OHaiDanny, line 325: @nonowazu Question on tier name field. Do you want me to list out all the values and put the in a dropdown or we get the content team to manually type it in?

exampleSite/content/encounters/uwu.md
OHaiDanny: Yeah I think making this optional for now would be good since we are only showing individual stuff at the moment

exampleSite/content/encounters/uwu.md
```diff
@@ -0,0 +1,15 @@
+---
+title: The Weapon's Refrain
```
nonowazu, line 2: We'll probably need to make a few changes in the editor here, but it's not so critical that I would block right now:

1. The title should be the actual full title of the fight (so that `.Title` things can pick it up for SEO)
2. We'll probably need to have a "short name" which will be the file name/route fragement name (see [here](https://github.com/The-Balance-FFXIV/balance-static/blob/8c603d2545795e5ae5a71fa613d1a13cae481657/static/admin/config.yml#L48) for how to override the title to a name)

exampleSite/content/encounters/uwu.md
nonowazu: Also, while I'm doing re-review, we probably need to get FE to generate a default image here so we can default it in hugo.

exampleSite/content/encounters/uwu.md
```diff
@@ -0,0 +1,15 @@
...
+card_image: /theme-assets/gaia-test.png
+tier_image: /theme-assets/gaia-test.png
+encounter_category: ultimate
+tier_name: Ultimate
+series_name: Ultimate
+weight: 2
+tier_weight: 1
+coming_soon: false
+spoiler: false
+expansion: Stormblood
```
nonowazu, line 12: Maybe this is a personal stylistic nit, but to save us from typing, we could probably abbreviate everything as `arr`, `hw`, `sb`, `shb` and `ew` and just expand when we hit frontend things.

exampleSite/static/admin/config.yml
nonowazu: I'd say manually type it in. Providing a list means we have to manually change the editor each and every time a tier gets added.

layouts/partials/cards/encounters/fight_card.html
```diff
@@ -0,0 +1,25 @@
+<div class="relative mb-8 overflow-hidden">
+  {{ if or (eq .Params.coming_soon true) (eq .Params.spoiler true)}}
+    <a class="block bg-card-light">
+  {{ else }}
+    <a class="block bg-card-light" href="{{ .Permalink }}">
+  {{ end }}
```
nonowazu, line 6: I don't think we want to remove the `href` if it's just a spoiler — only if it's coming soon

Also I feel like this is a super dumb nit from me, but:

```html
<a class="block bg-card-light"> {{ if (eq .Params.coming_soon false) }}href="{{ .Permalink }}"{{ end }}>
```

layouts/partials/encounter/homepage.html
```diff
@@ -0,0 +1,57 @@
...
+        <div class="grid grid-cols-3 gap-8 mb-8">
+          {{ range sort .Pages "Weight" "desc" }}
+            {{ partial "cards/encounters/fight_card.html" . }}
+          {{ end }}
+        </div>
+      {{ end }}
+    </div>
+  </div>
+  {{ end }}
+</div>
```
nonowazu, line 57: Not a now thing, but would it be good to make a partial/component that will just render a bunch of cards given a list? It might cut down on some of the repetitive stuff here.

Either way, just morning Nono thoughts.

nonowazu (changes requested):
I would remove the `gaia-test` from the theme assets, and we should probably replace it with something that's just a default image. For the rest of it, I'd like to pull the PR down and do a few tests with it, so I might have some more feedback later.

layouts/partials/cards/encounters/fight_card.html
OHaiDanny: My editor is colouring it weirdly but it works 😛

exampleSite/static/admin/config.yml
```diff
@@ -307,23 +307,46 @@ collections:
...
+  - label: "Encounters"
+    name: "encounter-data"
+    folder: "content/encounters"
+    identifier_field: "fight_title"
     create: true
     fields:
-      - {label: "Category", name: "category", widget: "string"}
-      - label: "Encounters" 
-        name: "encounters"
+      - {label: "Fight Title", name: "fight_title", widget: "string"}
```
nonowazu, line 317: I'd add a title field back above this, so we can use `.Title` elsewhere. 

Also, for this widget, might be best if we add a description *and* validation, so probably break this out like this:

```yml
fields:
  - label: Fight Title
    name: fight_title
    description: This is the short name of the fight, like "o12s" or "e12s-p2"
    widget: string
    pattern: ["[A-Za-z0-9\-]", "Cannot use spaces. Please only use letters, numbers, and dashes"]
```

layouts/partials/cards/encounters/fight_card.html
```diff
@@ -0,0 +1,21 @@
+<div class="relative mb-8 overflow-hidden">
+  <a class="block bg-card-light" {{ if (eq .Params.coming_soon false) }}href="{{ .Permalink }}"{{ end }}>
+    {{ if eq .Params.spoiler true}}
+      <img class="block object-cover w-full filter blur-md" src="{{ .Params.tier_image}}" alt="" />
+    {{ else }}
+      <img class="block object-cover w-full" src="{{ .Params.tier_image}}" alt="" />
+    {{ end }}
+    <div class="bg-card-light">
+      <h2 class="flex items-center text-md md:text-md uppercase px-4 font-semibold tracking-wider py-2 mb-0">{{ .Params.fight_title }}</h2>
```
nonowazu, line 9: I actually think this might be better to do as just `{{ .Title }}` so we get a meaningful name instead of `e12s` and instead get the good fight name.

layouts/partials/encounter/single_header.html
```diff
@@ -0,0 +1,31 @@
...
+      <div class="col-span-12 lg:col-span-9 text-white responsive-container">
+        <div class="flex items-center mb-7">
+          <a href="/encounters" class="flex flex-row items-center" >
+            <div class="card-title filter drop-shadow-lg-encounter">{{ .Params.series_name }} &raquo; {{ .Params.tier_name }}</div> 
+          </a>
+        </div>
+        <div>
+          <div class="flex items-center mb-4">
+            <img class="inline mr-8 filter drop-shadow-lg-encounter" src="/theme-assets/encounters.svg" />
+            <div class="font-bold filter drop-shadow-lg-encounter text-2xl lg:text-6xl uppercase">{{ .Params.fight_title }}</div>
```
nonowazu, line 13: Ditto with changing this back to `{{ .Title }}`

nonowazu (changes requested):
Small nits, but I think we're close

layouts/partials/encounter/single_header.html
OHaiDanny: Resolved!

layouts/partials/cards/encounters/fight_card.html
OHaiDanny: Resolved!

nonowazu (approved):

Issue #20, opened 2021-09-01
Implement search bar


nonowazu:
Likely using something like [hugo-lunr](https://www.npmjs.com/package/hugo-lunr), perhaps?

PR #243, opened 2021-11-30
Basic config for generating a search index for use with lunr

nonowazu:
This is only a first pass – I fully expect this logic to get *way* more complicated as time goes on, since we'll be making fields based on what type of page it is.


nonowazu:
This is also connected to #20

ArcaneDisgea:
One thing to consider is using the tags from the OpenGraph meta to populate a tags field within search. Using tags properly will not only increase SEO on elgoog but can also make our in-house search on the site more effective.

PR #248, opened 2021-12-15
Change up bis to support displaying etro sets better

nonowazu:
This PR (in no particular order):

* fixes a bug with heading anchors (the bug being that they didn't exist)
* changes the format from the link to the etro set being in the h1 to using our etro partial to embed the set (much cleaner IMO)
* Changes the *name*, but not the frontmatter param name (though we probably should, but what if we wanted to support, say, etro and a raw link?)

Either way, waiting on feedback.


nonowazu:
@hintxiv is there any intention of merging this PR? I wanted to give a chance to merge this before I deleted my local forks.

PR #261, opened 2022-02-02
Fix malformed google font link

22a:
The production site is currently attempting to load `Kumbh Sans` and `Roboto` font faces from google fonts, the request for which is failing:

Screenshot 2022-02-02 at 01 09 00

this is because the URL is formatted incorrectly:
```
400: Invalid selector

Too many values in tuple

Kumbh Sans:wght@100,700
                    ^
```

### This PR:
* Updates the font link to one with correctly formatted query params, which results in:
  * `Kumbh Sans` @ 100
  * `Kumbh Sans` @ 700
  * `Roboto` @ 100
  * `Roboto` @ 400
  * `Roboto` @ 700
  * `Roboto Italic` @ 100
  * `Roboto Italic` @ 400
  * `Roboto Italic` @ 700
    * I'm not exactly certain what this url was trying to import before but this feels close enough, and due to how font faces work we don't pay the cost of loading them if they're not used in the css for the page

Screenshot 2022-02-02 at 01 10 41

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -17,7 +17,7 @@
 <title>{{ .Site.Title }}{{ with .Title }} | {{ . }}{{ end }}</title>
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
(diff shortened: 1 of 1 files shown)


hintxiv:
lgtm, thanks for the PR!

hintxiv (approved):

PR #266, opened 2022-04-18
Add meta keywords to head partial

hintxiv:
Will require a corresponding CMS change in balance-static.


lailleo:
Looks good to me

lailleo (approved):
Looks good

PR #268, opened 2022-04-19
Feature: gsheets integration

rexatlanti:
This allows for the functionality of gsheets being added directly on the site. This will neccessitate a follow up PR for the CMS in balance static in order to be able to use it. Looks like this.

grafik


hintxiv (approved):
lgtm

PR #271, opened 2022-04-22
Add image modals

hintxiv:
VjjuQq0n (1)

Some pages will need to be adjusted to get rid of the image url link workaround.


rexatlanti (commented):
I have literally no idea about css but the rest looks sane?

rexatlanti (approved):

PR #278, opened 2022-06-24
move js to bottom of page

### LINKED ISSUE #218: Move all Javascript to the bottom of the page
From Nono:

This is not a 'now' thing, but a "Someone please remind me later" thing — we should move all javascript to the bottom of the page so that they're not blocking on load. We'll need a partial called "extraScripts" or something. Either way, someone please remind me later when we're doing a polishing pass.

nonowazu:
Addresses / fixes #218 

Might be worth discussion if this breaks anything. I don't think it does, but verify anyway.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -1,20 +1,27 @@
 <!DOCTYPE html>
 <html lang="en">
-    <head>
-        {{- partial "head.html" . -}}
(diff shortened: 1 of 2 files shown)


Hughes89 (approved):

lailleo (approved):
Approved per discussion in dev channel.

PR #279, opened 2022-06-25
proof of concept accurate previews for job landing

nonowazu:
PostCSS is back on the menu.

Though my superpower of "[reading the documentation](https://www.netlifycms.org/docs/customization/#registerpreviewtemplate)", I enabled "accurate" previews of content in the editor:

Screen Shot 2022-06-24 at 10 56 14 PM

As you can see from the single file modified, this allows you to completely reuse main style CSS and have the preview be representative of something that will land on the production site, and it will track correctly as the theme updates.

What I need from the reviewer:
1. List of page types to prioritize throughout the lifetime of this PR
2. Additional data to represent in the preview (header image, etc)
3. Feedback on where to split out all the previews to (I refuse to let them live in a script tag long-term)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Glam is a theme for The Balance website put together by the wonderful FFXIV comm
 
 ## Testing the theme locally
 
-Make sure you have [hugo](https://gohugo.io/getting-started/installing/) installed, either drop `hugo.exe` in the root dir or add it to your PATH.
+Make sure you have [hugo](https://gohugo.io/getting-started/installing/) installed, either drop `hugo.exe` in the root dir or add it to your PATH. Make sure you're using hugo version [v0.86.1](https://github.com/gohugoio/hugo/releases/tag/v0.86.1)
 
 ```sh
 # Clone the repository
@@ -14,8 +14,8 @@ git clone https://github.com/The-Balance-FFXIV/glam.git
 cd ./glam/
 
 # Install project dependencies and dependencies. If you prefer to not edit PATH, you can install globally.
-export PATH=$PATH:node_modules/.bin
-yarn install
+export PATH=$PATH:${PWD}/node_modules/.bin
+yarn
 
 # Run the servers
 yarn start:static
@@ -23,13 +23,13 @@ yarn start:static
 # Server starts by default on `localhost:1313`
 ```
 
-
 ## Using the editor locally
 
 First run the static site server with `yarn start:static`, then run `yarn start:admin` and navigate to `http://localhost:1313/admin`.
 Alternatively, `yarn start` will start both servers.
 
 # The wiggly bits
+
 This is some general documentation until we can clean it out.
 
 ## Adding an item to the footer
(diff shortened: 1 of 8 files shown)


lailleo:
I'll see if we have someone with knowledge required to look at this -- most of the other devs needed to step away for work reasons or otherwise in the last few months. Lemme post it up there and see.

rexatlanti:
1. The general "guide" pages should probably be first. Idk what we should do after. Maybe encounter tips.
2. as you've said, header image would be neat. In general images being the size they are on the live site. Yt stuff would be cool if it's possible to render the shortcodes in the preview. If not then idk if it's possible to grab a placeholder still image to put there via some sort of api? Or even just a static placeholder thing.
3. as by your suggestion, `glam/static/theme-assets/editor/job-preview.js` is as fine as any place to me.

nonowazu:
Some more progress. I broke out my one test case to it's own file, which is a start

https://user-images.githubusercontent.com/83298827/175791896-cf9567a8-fd9b-44f1-87b5-404b4dfe5093.mov

I also realize after working with the data in the yaml, we're only going to get the username fields, which means no rendering in-preview for posts, but it also means it can do a somewhat nice rendering in the authors section I guess.

nonowazu:
By way of update, here's what the preview looks like for the following guide types:
* Job Leveling Guide
* Job Basic Guide
* Job Skills Overview
* Job Openers

image

I'm going to see if I can't be clever about the faq/bis parts for reusable "components", but I may also "meh" my way through it. I would really appreciate it if someone else cloned this branch down and tried it out and waxed poetic (read: give input) about how to render other components such as the header image, since I am not a styling wizard.

lailleo:
I'll toss this in the dev channels and see if we have any devs available to give some input. Appreciated, as usual.

ScrubbyXIV:
@nonowazu the PR looks fine to me, I'm not quite sure how much you really can reuse of the job guide compared to other pages. I'm sure there's some way, as it seems you already had the idea with the authors for that.

nonowazu:
@ScrubbyXIV - the biggest reuse would be rendering the "container" divs - since that should be the same on 90 percent of 'article' pages (fight guides, job guides, articles, news posts, etc).

ScrubbyXIV:
@nonowazu then I'd suggest to move the render container function and the author list into a separate file and have generic classes, that way we can properly reuse them for any sort of guides, like the job one.

nonowazu:
@ScrubbyXIV there are several reasons I've kept this to one file:

1. the code looks messy since you're depending on people importing them successively
2. I can't take advantage of build systems like webpack *et al* (or I'd be writing react code instead of the `h` render function mess)
3. There's a whole mess of js building "features" in the hugo docs [here](https://gohugo.io/hugo-pipes/js/), but I was hoping someone else would read those docs and either PR to my branch or otherwise let this PR merge before starting "functional refactoring"

ScrubbyXIV (approved):

lailleo:
apropros of absolutely nothing, I tried building this locally, but keep getting:

"Error: Error building site: POSTCSS: failed to transform "css/main.css" (text/css): Error: Loading PostCSS Plugin failed: Invalid or unexpected token"

Any ideas so I can actually try to look these over locally going forward?

PR #299, opened 2022-09-19
Change the admin editor from a static affair to a dynamic one

nonowazu:
I'm pushing this up as a draft because I'm thinking about how best to implement the `.yml` file section as front matter params or hard-coded or something else (feedback welcome).

This changes the editor from `/static/admin/index.html` to `/content/admin.md` - the reason for this change is because PR #279 added previews, but they don't work with a minified and fingerprinted css. This PR addresses that (and, in theory, lets you change the route to the editor by just moving that `.md` file around).


nonowazu:
This should be ready for review now. For information on how `newScratch` works, see docs [here](https://gohugo.io/functions/scratch/#the-local-newscratch); as you might know, if statements and etc are kind of their own context and you can't do something like this python equivalent:

```python
foo = 'some default'
if DEVELOPMENT:
    foo = 'some dev default'
```

Because the inner `foo` is lost in hugo-land. So I landed on scratch to solve this issue.

hintxiv:
I like this solution for grabbing the css, but I'm not sure if it's practical to use it for the Netlify config since there are only a handful of lines different between `config.yml` and `dev.yml` vs. hundreds of lines that would need to be duplicated. Is there a way around having to duplicate changes in both files?

nonowazu:
You could render the `.yml` files using other [output formats](https://gohugo.io/templates/output-formats/), but this seemed like a nicer medium-term solution.

I'd be willing to offer a compromise and get this for now, and either offer dynamic config rendering (because I know none of you want to actually touch the config file – I don't either) or leave as stopgap until phase 2, which I actually have progress on (at least from the client and server side, but the code requires some cleanup to be fully functional)

hintxiv (approved):
Let's get this in 👍

Issue #285, opened 2022-08-09
Limit (and expand) screenshot generation workflow


nonowazu:
By messing with the `on` section in the workflow, you could limit the workflow to only ones that affect the front page (see [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths)).

At the same time, this provides an opportunity to clean up the workflow to show any of the different layout types on an appropriate PR change.

PR #300, opened 2022-09-20
Stop GHA cluttering up things when it isn't needed

nonowazu:
What is says on the tin. Related to #285, this is a non-exhaustive list of things that only affect the front page. Another step would be to tie certain layouts to certain screenshot generation workflows, but that's a PR for another time – my goal right now is to stop the annoying workflow when we aren't changing anything.


rexatlanti (approved):
lgtm

Issue #69, opened 2021-11-06
Layout for `/jobs/` route


nonowazu:
Since we're already using the job layout's `list.html` and `single.html` for paths like `/jobs/$role/$job` paths, the `/job` path looks just like a job landing page:
<img width="1663" alt="Screen Shot 2021-11-06 at 2 24 30 PM" src="https://user-images.githubusercontent.com/83298827/140621504-0342a369-2ce9-40d5-8b66-352efcc11343.png">

PR #301, opened 2022-09-20
Add client-side redirection to the theme

hannachen:
https://github.com/The-Balance-FFXIV/glam/issues/69 nice ( ͡° ͜ʖ ͡°)

Kind of cheesed... not quite a 301 Redirect as GH Pages only host static pages and [does not support server side](https://stackoverflow.com/questions/70875402/how-to-render-dynamic-pages-on-github-pages) redirects 😓

~Add a new layout type `redirect` to perform a client-side redirect.~
Add client-side redirection markup to the base template and introduce a partial to render the "redirecting..." message.

## Usage

In a content page, assign the type to `redirect` and set param `redirect_to` to the path or URL you would like to redirect to.
Optionally, add a `redirect_after` in **seconds** to wait before redirecting.

```markdown
---
title: Page Title
type: redirect
redirect_to: /
redirect_after: 3
---

My page content
```

If a delay is defined, this is what the visitor will see during the wait:
Screen Shot 2022-09-20 at 6 57 18 PM


nonowazu:
The biggest thing worth discussing here is – is there a plan to have `/jobs/` be its own page at some point as a landing page for generalized articles about jobs? If there's a 301 redirect, that's a pretty [harsh](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) solution (hence the "permanently" part of the error code), and potentially bad for SEO if it ever got that far.

hintxiv:
Thanks for the PR, and nice work getting the PR with 301 in the title as #301 :)

The `redirect` layout seems fine to me but I'll wait for @nonowazu to weigh in before merging.

nonowazu:
My only holdover is that the 301 is a permanent redirect. Could we do a [302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) instead?

Alternatively, if you really wanted to make this redirect layout "feature complete", you could have some front matter to customize part(s) of the redirect (would be useful if you ever decided to do https://thebalanceffxiv.com/links/foo/ as an ad-hoc linking service)

layouts/redirect/list.html
```diff
@@ -0,0 +1,14 @@
...
+<html lang="en">
+  <head>
+    <meta http-equiv="refresh" content="0; url={{ $redirect_to }}">
+    <link rel="canonical" href="{{ $redirect_to }}">
+    <script>
+      window.location.href = "{{ $redirect_to }}";
+    </script>
+  </head>
+  <body>
+    <p>Please click this [link]({{ .Params.redirect }}) if you are not redirected.</p>
```
nonowazu, line 12: Since this isn't markdown - just pure html, you'd need to change this to

```html
<p>Please click this <a href="{{ $redirect }}">link</a> if you are not redirected.</p>

**Review by nonowazu — CHANGES_REQUESTED** [review]

**Comment by nonowazu** [comment]
Still researching, but there doesn't appear to be a clean way to do a 302 redirect with just html like this.. Maybe it's worth building out a job landing page (just like the role landing page – we could even take a good number of the components for this)

**Comment by hannachen** [comment]
> Thanks for the PR, and nice work getting the PR with 301 in the title as #301 :)

TY sounds like I should go buy a lottery ticket!

> Still researching, but there doesn't appear to be a clean way to do a 302 redirect with just html like this.. Maybe it's worth building out a job landing page (just like the role landing page – we could even take a good number of the components for this)

I'm not sure it's possible to perform a server side redirect with 301 or 302 status code when hosting with GitHub Pages, status codes can only come from the server side.  I would love to learn how to customize server status codes on GH Pages if you find it! The beauty of this approach is that it's a temporary client-side redirect :)

I'm not an expert on SEO but a quick search suggest that old school meta refresh redirects [shouldn't negatively impact SEO](https://www.contentkingapp.com/academy/redirects/faq/html-meta-redirect-bad-seo/).

So I was thinking we can create an `_index.md` file under `/content/jobs/` in the static site project to handle the redirect (as demonstrated in `exampleSite`) until a job landing page is ready to go. To restore it from a redirect to a regular page, change the `type` field and remove `redirect_to`.

If we want to keep this template for the long-term, it can be used to handle other redirects. This method of redirection can be removed once the site can be hosted elsewhere with more server side control.

For making this "feature complete", do we have examples of common customization use case? Maybe tracking? 🤔 Considering that server-side redirects probably performs better, I would count this as a workaround to GitHub Pages hosting.

Also, good eye, ty for noticing the markdown instead of HTML, will fix accordingly!

**Comment by nonowazu** [comment]
For 'feature complete' bonus points, a customizable (but default if not supplied) time for redirect might be important. I don't predict many people would customize, but might be useful if an issue is encountered down the road. It might also be worth to not override baseof, but instead add to the headers and body directly, so if it gets stuck on the page, at least it's not a blank white page with some html – however, that's so minor that I'm loathe to really bring it up.

I would love to keep the redirect template for other things, though, just because you might have a google sheet and it would be nice to do something like `https://thebalanceffxiv.com/links/blmsheet/` and just have it magically redirect or something.

My concern over the redirect might not be for SEO, but for browsers that cache the route. I guess it's not really worth thinking too hard about if even one other person thinks it's okay.

File: layouts/redirect/list.html [hunk 2]
```diff
@@ -0,0 +1,14 @@
+{{ $redirect_to := .Params.redirect_to }}
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta http-equiv="refresh" content="0; url={{ $redirect_to }}">
+    <link rel="canonical" href="{{ $redirect_to }}">
+    <script>
+      window.location.href = "{{ $redirect_to }}";
```
nonowazu, line 8: A minor nit, if we're doing time setup, might as well do a timeout in the head to the same amount of seconds as the refresh meta tag.

hannachen:
> For 'feature complete' bonus points, a customizable (but default if not supplied) time for redirect might be important. I don't predict many people would customize, but might be useful if an issue is encountered down the road. It might also be worth to not override baseof, but instead add to the headers and body directly, so if it gets stuck on the page, at least it's not a blank white page with some html – however, that's so minor that I'm loathe to really bring it up.
> 
> I would love to keep the redirect template for other things, though, just because you might have a google sheet and it would be nice to do something like `https://thebalanceffxiv.com/links/blmsheet/` and just have it magically redirect or something.

Sounds good, I'll expand on this a bit more so that it's done in a less hacky way.

> My concern over the redirect might not be for SEO, but for browsers that cache the route. I guess it's not really worth thinking too hard about if even one other person thinks it's okay.

Luckily for us, it appears that GitHub Pages only instruct browsers to cache pages for [10 minutes](https://github.com/community/community/discussions/11884)! So I think we should be ok in that department :)

nonowazu:
I think between the time and GH caching for only 10 minutes this is a fine PR as-is.

hannachen:
Thanks for the feedback. I made some changes according to our discussion.

layouts/partials/redirect.html
```diff
@@ -0,0 +1,12 @@
+{{ if .Params.redirect_to }}
+  {{ $redirect_to := .Params.redirect_to }}
+  {{ $redirect_after := .Params.redirect_after | default "0" }}
+  <meta http-equiv="refresh" content="{{ $redirect_after }}; url={{ $redirect_to }}">
+  <link rel="canonical" href="{{ $redirect_to }}">
+  <script>
+    var redirectTimeout = parseInt("{{ $redirect_after }}") * 1000;
```
nonowazu, line 7: Since this is a number, wouldn't the substitution look like this (for, say, 3):

```js
var redirectTimeout = parseInt("3") * 1000;
```

Wouldn't it be better to do something like:

```js
var redirectTimeout = {{ $redirect_after }} * 1000;
```

Which would work out to `var redirectTimeout = 3 * 1000;` (reusing my 3 second timeout example)

layouts/partials/redirect.html
```diff
@@ -0,0 +1,12 @@
+{{ if .Params.redirect_to }}
+  {{ $redirect_to := .Params.redirect_to }}
+  {{ $redirect_after := .Params.redirect_after | default "0" }}
```
nonowazu, line 3: Do we want a default of 0 here? Or do we want a default of something like 1? I only ask because it might be beneficial to let it sit for at least a second.

layouts/_default/baseof.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     {{- partial "head.html" . -}}
+    {{- partial "redirect.html" . -}}
```
nonowazu, line 5: I'll be honest that I think I preferred it when redirect was its own layout instead of directly being inserted into `_default`, because this feels like a gun pointed at our feet.

With it being its own, at least you'd have to purposefully select the layout with `type` in the front matter. Sorry to make you go back in work on this, but I think this is an unsafe solution.

nonowazu (changes requested):

layouts/_default/baseof.html
hannachen: > It might also be worth to not override baseof, but instead add to the headers and body directly, so if it gets stuck on the page, at least it's not a blank white page with some html

Based on the comment above from earlier, I was under the impression that we do not want to override the base template. I can change it back but I would like to confirm that we _do_ want this inside a separate template with a similar layout.

layouts/partials/redirect.html
hannachen: I don't feel strongly either way, but I am leaning towards 0 as a default because I expect redirects to be immediate, unless there is a good reason to let the page flash for a second.

layouts/partials/redirect.html
hannachen: I think your suggestion is more ideal if it works! I am unfamiliar with this templating language and using the variable without quotes showed an error in my IDE, will give this a try!

hannachen: It works! But VSCode is angy even after I installed an [extension](https://marketplace.visualstudio.com/items?itemName=budparr.language-hugo-vscode) to support HUGO syntax highlighting. Any suggestions? 
Screen Shot 2022-09-21 at 6 47 20 PM

nonowazu: I've not been happy with the highlighting situation in hugo in tools like vscode because of edge cases like this, and I've hit my own edge cases where the highlighting goes crazy; but the code is *technically* correct since it's hugo -> html, and the final output matters the most in this case.

Also, when you started using vscode with the repo, did it recommend some extensions? Because that extension is one of the [recommended](https://github.com/The-Balance-FFXIV/glam/blob/e70e11d8a82f46115d0fcb99ace79e4691daec2b/.vscode/extensions.json#L3) ones anyway.

layouts/partials/redirect.html
nonowazu: Well, I'm thinking about the use case of redirecting to other external things as a tool (and that's good because people aren't just jolted out of the site suddenly); I feel like the only case where it might be used to redirect internally would be `/jobs/` -> `/`, so that's the minority case.

layouts/_default/baseof.html
nonowazu: I worded that poorly; I was implying that we should *duplicate* `baseof`, but re-reading myself I did not make it clear. I think in 99% of the cases that overriding the base html template is probably bad, except in cases like:

1. It fundamentally changes how the page works (in your case, making it so that a single front-matter piece makes the page change from a static page affair into a redirect)
2. It requires the layout/style/etc to be so different that it would be a dirty hack to make work (in [my](https://github.com/The-Balance-FFXIV/glam/pull/299) case with the dynamically generated editor bits)

nonowazu (approved):
All of these changes look good! Thank you for putting up with my nitpicking!

layouts/partials/redirect.html
hannachen: I think for the purpose of simply not exposing `/jobs/`, we should keep this as a fast internal redirect. Perhaps we can make changes in the future to repurpose this into something that is meant to handle external links, maybe with better copy and UI like adding animation to countdown the timer.

hannachen:
Thanks! For some reason this PR is still showing "Review required" for me and I'm unable to merge. So if possible, please merge this when you can.

hintxiv:
> Thanks! For some reason this PR is still showing "Review required" for me and I'm unable to merge. So if possible, please merge this when you can.
> 
> Edit: or does that approval have to come from me? I've never merged PRs across forked repos 😓

A maintainer has to be the one to merge it, which unfortunately is me in this instance. Feel free to ping / DM me on [Discord](https://discord.com/users/492781599126061066) to get a quicker response. Thanks again for the PR!

hintxiv (approved):

PR #302, opened 2022-09-20
Remove reference to Hugo v0.86.1 in README

hintxiv:
The latest versions of Hugo are now working as of #282


nonowazu (approved):
We already mention the 0.101 in the readme, so this is a good removal.

lailleo (approved):
Looks fine to me as well

PR #303, opened 2022-09-21
SEO description and tags in editor

nonowazu:
I'm not happy with how the widgets are working out so far, but it's a start.

It may be good to have a "data" section for "approved tags" (the same sort of thing done with user profiles), so they're easy to select. I also haven't added images to that, so conversation/feedback is important prior to landing this PR.


lailleo:
I would agree with adding a data section with approved tags to make things more uniform as opposed to everyone choosing their own, yeah.

nonowazu:
Depending on how #308 works out, this might have to be closed in favor of expanding out the dynamic generation component.

PR #305, opened 2022-09-26
make glam work as a development package and as an included theme

nonowazu:
Basically, `yarn start:dev` will start work with the theme in `exampleSite`, and `yarn start` will work with anything where glam is included as a theme.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "scripts": {
     "start": "concurrently --kill-others-on-fail --names static,admin \"yarn start:static\" \"yarn start:admin\"",
     "start:dev": "cross-env GIT_REPO_DIRECTORY=exampleSite concurrently --kill-others-on-fail --names static,admin \"yarn start:static:dev\" \"yarn start:admin\"",
-    "start:static": "hugo server -s exampleSite/ --themesDir=../.. --disableFastRender",
+    "start:static": "hugo server --disableFastRender",
     "start:static:dev": "hugo server -s exampleSite/ --themesDir=../.. --disableFastRender",
     "start:admin": "node_modules/.bin/netlify-cms-proxy-server"
   }
--- a/package.json
+++ b/package.json
@@ -23,8 +23,9 @@
   },
   "scripts": {
     "start": "concurrently --kill-others-on-fail --names static,admin \"yarn start:static\" \"yarn start:admin\"",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,7 +23,8 @@
   },
   "scripts": {
     "start": "concurrently --kill-others-on-fail --names static,admin \"yarn start:static\" \"yarn start:admin\"",
+    "start:dev": "cross-env GIT_REPO_DIRECTORY=exampleSite concurrently --kill-others-on-fail --names static,admin \"yarn start:static\" \"yarn start:admin\"",
     "start:static": "hugo server -s exampleSite/ --themesDir=../.. --disableFastRender",
```
hintxiv, line 27: This bit looks for `exampleSite/` which wouldn't exist outside of the theme.

hintxiv (changes requested):

package.json
nonowazu: The `GIT_REPO_DIRECTORY` environment variable is propagated down to the `start:static` invocation, which then uses that variable (per the [docs](https://github.com/netlify/netlify-cms/tree/master/packages/netlify-cms-proxy-server#custom-configuration))  - at least it did on OSX; this seemed cleaner than duplicating both `start:static` and `start:admin` scripts

hintxiv: I think `GIT_REPO_DIRECTORY` only applies to netlify, hugo itself is being given `-s exampleSite/` which does not exist outside of the theme directory. Here's the error I get (I'm on Linux) when I try to run `start` from within `balance-static`:

image

nonowazu: Are you running this as part of balance-static, or part of glam? I guess I can hook it up in a test balance-static branch or something.

hintxiv: As part of `balance-static`, `start:dev` looks like it works fine.

nonowazu: check again with the latest update; no errors on my side(but no guarantee the admin works because it doesn't have the dynamic admin changes or dev.yaml yet)

lailleo:
Working fine for me on Windows.

lailleo:
Putting out one last call for a follow up on this but if nothing, I'm going to merge this tomorrow.

hintxiv (changes requested):
Running `yarn start` from the outer directory still doesn't work.

lailleo:
image

nonowazu:
It's been fixed, but without changes from other PRs it's going to fall flat. I did test that `yarn start:dev` now works as expected. balance-static also works, but relies on configuration changes, which I have work happening [here](https://github.com/nonowazu/netlify-config-generator) along with some code to commit here later.

hintxiv (approved):

PR #308, opened 2022-10-19
Generate config dynamically

nonowazu:
Add some python and a small library I created to dynamically generate the netlify configuration. Successful dogfooding of this in dev should result in me pushing that library to pypi for future consumption (removing the git dependency in requirements.txt).

Current instructions for use (from the scripts directory):

* (optional) Make a virtualenv with `python -mvenv env` (I think this works with python 3.10 or higher, but when I'm done it should be 3.7 or higher)
* `pip install -r requirements.txt`
* `python generate_config.py > ../exampleSite/static/admin/dev.yml`
* test with usual `yarn start:dev` commands

I wouldn't bother too hard with reviewing dev.yml, but I figured it would be nice to include it as part of this PR for "rapid start" (I don't see this changing too often)

### DIFF
--- a/exampleSite/content/jobs/healers/scholar/_index.md
+++ b/exampleSite/content/jobs/healers/scholar/_index.md
@@ -6,7 +6,7 @@ menu:
     identifier: scholar
     name: Scholar
     parent: healers
-seo-description: The landing page for Scholars
+description: The landing page for scholars, where all the cool kids hang out
 tags:
   - FFXIV
 ---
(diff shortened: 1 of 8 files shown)


hintxiv:
Code looks good, could you put the instructions for running the script somewhere that people will be able to find them? (Also maybe a package script if possible?)

nonowazu:
I put the instructions in `scripts/README.md` which should be pretty immediately available when browsing github on the web, or just opening and reading.

hintxiv (approved):

PR #314, opened 2023-06-19
Add missing font-weights for Roboto

teatwig:
The font weight 300 was missing, which made "font-light" fallback to the font weight 100. This provides bad readability at a font size of 14px.
Roboto does not provide the font weight 600 for "font-semibold", but for consistency I added the font weight 500 as well.
I also updated the admin index.html so the fonts used in the admin UI match those in the theme. It might make sense to update it in the main repository as well.

This has been especially visible on the copyright text on the footer.

Before:
image

After:
image

### DIFF
--- a/layouts/editor/baseof.html
+++ b/layouts/editor/baseof.html
@@ -17,7 +17,7 @@
     <link rel="preconnect" href="https://fonts.googleapis.com" />
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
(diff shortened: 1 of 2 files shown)


teatwig:
It probably make sense to create a partial with tags inside head that are shared between the theme and admin UI. 
Although the question is if the admin UI even needs these changes. I haven't looked that deeply into the features of Netlify CMS.

lailleo (approved):
Approving per previous feedback.

PR #315, opened 2024-01-20
Update what should be staging to match prod

nonowazu:
Looks like prod's github workflows, hugo version, etc have been updated a bit. This PR brings a few foundational things in line with those changes so that future PRs will be more 1:1


furstblumier (approved):
lgtm

PR #326, opened 2024-01-20
add .github to the paths that are checked

furstblumier:


nonowazu (approved):
Looks good to me; we can use this to create more appropriate linting/preview PRs down the road

PR #330, opened 2024-01-21
ignore license and .md files

furstblumier:


nonowazu (approved):
Last one for now. Every new one for the next 8 hours will cost at least a quarter.

PR #347, opened 2024-05-24
First pass at a more full-featured configuration generation

nonowazu:
This is a more full-featured version of my previous efforts to generate the configuration file with the idea that adding new editor pages to the netlify configuration should be driven by a more model-driven approach. I invite anyone to give me feedback about this in any area since this is a lot.


tools/econfgen/econfgen.yml.example
```diff
@@ -0,0 +1,44 @@
...
+    - ['Sage', 'sge']
+    melee:
+    - ['Monk', 'mnk']
+    - ['Dragoon', 'drg']
+    - ['Ninja', 'nin']
+    - ['Samurai', 'sam']
+    - ['Reaper', 'rpr']
+    ranged:
+    - ['Bard', 'brd']
+    - ['Machinst', 'mch']
```
Snipa22, line 30: Spelling error here for MCH, should be Machinist based on current generated yaml, catches if you just copy the example out.

Snipa22:
Verified diff generated between scripts/generate_config.py and econfgen, the econfgen version adds `Fight-specific tips for <class>`, otherwise the output yaml is the same other than some site configuration variables using the econfgen.yaml.example as a base.

lailleo (approved):
Approving per prior review

PR #348, opened 2024-06-07
Add discord icon to discord username

sneakycrow:
To address The-Balance-FFXIV/balance-static#1001, adds a discord icon to discern what the username is for. 

I did notice that the original issue stated adding lighter text or the icon, and I can see lighter text was added. But, considering discord usernames no longer have their notorious `#123` suffix, I think when those start to be updated the distinction of it being a discord username will be valuable.

### DIFF
--- a/layouts/partials/components/author_box.html
+++ b/layouts/partials/components/author_box.html
@@ -6,6 +6,9 @@
   </div>
   <div class="text-center text-gray-light">{{ .name }}</div>
   {{ with .socials.discord_id }}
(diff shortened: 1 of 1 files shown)


nonowazu:
This is good stuff - thank you! Would it be possible to show a before/after of this change just for documentation's sake?

sneakycrow:
> This is good stuff - thank you! Would it be possible to show a before/after of this change just for documentation's sake?

Yes I can, here you go

Left Before, Right After

image

nonowazu (approved):

PR #349, opened 2024-06-11
Fix embarassing typo that keeps glam from working in production..

nonowazu:
Like it says; I'm a bit embarrassed to be honest.


sneakycrow (approved):

lailleo (approved):
image

PR #350, opened 2024-06-16
Add options that are missing from current production generation

nonowazu:
Using [`deep`](https://github.com/seperman/deepdiff) with the current production yaml in balance-static and the generated yaml in glam results in some discrepancies:

``` 
[3.12|env][king@guren:balance-static (main[✘!?])]$ deep diff --ignore-order static/admin.old/config.yml static/admin/config.yml | head -n 10
{
  "dictionary_item_added": [
    "root['local_backend']"
  ],
  "dictionary_item_removed": [
    "root['publish_mode']",
    "root['site_url']",
    "root['show_preview_links']",
    "root['backend']['site_domain']"
  ],
```

This PR adds almost all those fields in to the generation.

I'm not 100% sure that `site_domain` is [absolutely required](https://decapcms.org/docs/backends-overview/) (someone obviously double-check me here), and if it is it will require an update on my [upstream repo](https://github.com/nonowazu/netlify-config-generator/blob/main/netlifyconfig/netlify.py), but in the interest of speed I'll patch it locally if it becomes an issue.

### DIFF
--- a/tools/econfgen/econfgen.yml.example
+++ b/tools/econfgen/econfgen.yml.example
@@ -38,6 +38,9 @@ default:
 prod:
   # any items here will override everything in the default
   local_backend: false
+  publish_mode: editorial_workflow
+  site_url: https://glam.thebalanceffxiv.com/
+  show_preview_links: true
   backend: 
     name: github
     repo: foo/bar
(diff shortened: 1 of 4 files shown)


schmali18 (approved):

lailleo (approved):
Approving per prior review.

PR #351, opened 2024-06-18
Fix jumping cursor issue on editor in chrome

nonowazu:
Just like it says on the tin - this was fixed in balance-static previously, but I don't think the fix ever propagated back. Or I was blinded ~~by science~~ and missed it.


lailleo (approved):
Approving

PR #355, opened 2024-06-18
Add support for optional job intermediate guide page

Vantablak:
- Add optional Intermediate Guides section to list.html (`/job` page)
- Modify generate files to add intermediate guide section to the CMS
- Did not push changes to dev.yml/config.yml, as it seemed to include info about pct/vpr as well that wasn't already added, so unsure whether it needs to be committed
- Current layout of the job index page needs feedback, not sure if intermediate guides warrants a whole section.
- Might wanna change the icon for intermediate guides (currently using the same graduate cap as advanced guides)

image

### DIFF
--- a/exampleSite/content/jobs/melee/dragoon/advanced-guide.md
+++ b/exampleSite/content/jobs/melee/dragoon/advanced-guide.md
@@ -1,0 +1,16 @@
+---
+title: Dragoon Advanced Guide
+card_header_image: /img/jobs/drg/advanced.png
+authors:
+  - Balance-DRG-Staff
+patch: "6.2"
+lastmod: 2022-09-21T20:41:44.518Z
+changelog:
+  - date: 2021-11-15T21:03:02.264Z
+    message: Added page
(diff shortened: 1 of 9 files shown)


nonowazu (changes requested):
@Vantablak - can you use the editor in dev mode to add some example content so it's easily visible. It makes it easier to test that you've gone end to end with the editor changes.

Vantablak:
> @Vantablak - can you use the editor in dev mode to add some example content so it's easily visible. It makes it easier to test that you've gone end to end with the editor changes.

@nonowazu 
So just pushing the changes the dev.yml and exampleSite/job/blah/intermediate-guide.md is cool?

nonowazu:
Yes, the entire reason the `exampleSite` directory exists is to:

1. test the editor
2. have content that doesn't affect the theme while showing that it works without requiring the `balance-static` repo.

Also you don't have to, but it might be nice to run `rye run ruff format src` since I see I missed a newline.

nonowazu (approved):

PR #357, opened 2024-06-20
Fix viper svg and data to be consistent with other jobs

nonowazu:
Second verse, same as the first


nonowazu:
image

Also looks a little wonky, but we're on short time here.

PR #362, opened 2024-06-20
Add more preview styling for guide entries

nonowazu:
It was ~~thousands~~ two years ago when I added the preview styles and I forgot to add some for the guide types we had (and now have). The faq/bis ones still need to be written, so I'll put in a github issue to track against.


lailleo (approved):
Simple enough

PR #364, opened 2024-06-20
Remove old config generation; econfgen is the way

nonowazu:
The old one was a basic effort, but econfgen is being used in glam *and* balance-static now, so I think that clearly works better[.](https://www.youtube.com/watch?v=i7gIpuIVE3k)


lailleo (approved):
Approving

PR #379, opened 2024-07-25
Fix bis partials

xpdota:


layouts/jobs/bis.html
```diff
@@ -24,11 +24,8 @@
         <div class="markdown max-w-none">
           {{ range $index, $bis := .Params.bis }}
             <h2 id="{{ $index }}">{{ $bis.name }}</h2>
-            {{ if eq $bis.type "etro"}}
-              {{ partial "components/etro.html" $bis.link }}
-            {{ end }}
-            {{ if eq $bis.type "gsheets"}}
-              {{ partial "components/gsheets.html" $bis.link }}
+            {{ if templateExists (printf "partials/components/%s.html" $bis.name) }}
```
nonowazu, line 27: Per our discussion last night, I'm re-linking the [documentation](https://gohugo.io/functions/templates/exists/)
There are several problems here:

1. It should be `templates.Exists` (plural template, and a `.` between the two words)
2. It would be better to set up a variable on the previous line so you can reuse it:
```
{{ $partialPath := printf "components/%s.html" $bis.type }}
```
3. You should be using `$bis.type`, not `$bis.name`
4. Your partial is named `xivgear.html`, but you set the editor config to set the type as `xivgearset` (per #327); you need to fix either the name of the file in `layouts/partials/components` or amend config generation in `tools/econfgen/src/econfgen/fields.py`
5. If we're going to make embeds something that gets detected, I'd really like you to `mkdir layouts/partials/components/embed` and move all the embedded items there - this opens the potential to freely use them in markdown later via shortcode and removes the potential error of someone finding a way to accidentally use the other (non-embedded) partials in that directory

nonowazu (changes requested):

PR #408, opened 2025-04-07
Update BiS partials for extensibility

amakiirs:
Reviving #379 in an effort to (a) clean up the BiS interface in general and (b) update to the new form of xivgear links.

I expect some changes will be needed --- in particular, ideally, we can find a way to not rely on full links for the xivgear partial, but at least for now, making it work in the same way as the generic iframe (but with a separate tag for later handling) is a start.


layouts/partials/components/embeds/genericimage.html
```diff
@@ -0,0 +1,6 @@
+<div class="h-96">
+  <img class="w-full h-full" 
+    src="{{ . }}"
+    alt="An embedded image">
```
nonowazu, line 4: The only (non-blocking) feedback I have is that this should likely be a self-closing tag.

nonowazu (dismissed):

layouts/partials/components/embeds/genericimage.html
amakiirs: Should be addressed in latest version.

nonowazu (approved):

PR #410, opened 2025-04-09
Update some data to account for new updates

nonowazu:
Wonderful description, but we had some embed and econfgen updates and I wanted to make sure the exampledata and dev.yml matched.

Manually review the dev.yml at your own peril.


amakiirs:
Looks like a reasonable reflection of the changes to me.

PR #413, opened 2025-08-16
fix editor preview not showing page styling on certain page previews

kromuluss:
fixes certain pages in the editor preview not using real website styling for BIS/stat-priority/job-changes/FAQ pages

the root problem of why it didn't style those pages seems to just be that it was never added in the first place where it was added for the other pages


VioletStardustXIV (approved):

nonowazu:
Not going to add a review because this is a general observation - this isn't a good idea. The template previews work on looking at the data that's part of the frontmatter (and body) and rendering those. The regular previews work for the rest because they have that data - these won't and will look incomplete (and technically lose information in preview). I had worked with Sila on trying to resolve that, but I do not have an idea what the status is on this.

PR #416, opened 2025-08-17
fix for using full etro links in editor preview

kromuluss:
the original implementation was only using the set ID for the link, and it wouldn't properly preview the set information when users pasted in a full link (we've encountered this problem before on the live site, but i figured it should be fixed for the previewer as well) and so i included something to regex the ID into the iframe src if it was a full etro link

also temporarily disables the 2560px responsive container breakpoint until the BIS toc is fixed (because its just too wide) and widens the requirement for the 1920px breakpoint to 1930px so that it doesn't show up on fullscreen 1920x1080p browser windows because it widens too much

image

### DIFF
--- a/static/theme-assets/editor/job-preview.js
+++ b/static/theme-assets/editor/job-preview.js
@@ -23,32 +23,22 @@ const renderBisList = function (bis) {
         const name = h("h2", {}, bis.name);
         const type = bis.type;
         const description = h("p", {}, bis.description);
-        let link = bis.link;
         
-        switch(type) {  
-          case "etro":  
-            const etroLink = link.match(/etro.gg\/gearset\/([A-Za-z0-9-]+)(?:[?#]|$)/i);  
-            if(!etroLink) {  
-              link = `https://etro.gg/embed/gearset/${link}`  
-            }
-            break;  
-          default:  
-            break;  
-}  
+        let link = bis.link;
+        const etroLink = link.match(/etro.gg\/gearset\/([A-Za-z0-9-]+)(?:[?#]|$)/i);
+          if (type === "etro" && etroLink) {
+            link = etroLink[1];
+          }
 
-    // Handle special cases of the BiS display  
-    switch(type) {  
-      case "plain-text":  
-        // Plain text does not need an iframe, and can just be included as the given text (in the link field)  
-        var bisFrame = link;  
-        break;  
-      default:  
-        var bisFrame = h("div", { class: "h-96" }, h("iframe", {  
-          src: link,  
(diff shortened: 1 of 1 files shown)


static/theme-assets/editor/job-preview.js
```diff
@@ -22,13 +22,19 @@ const renderBisList = function (bis) {
...
-          type != "plaintext"
+          type != "plain-text"
             ? h("div", { class: "h-96" }, h("iframe", {
-                src: 
-                  type === "etro" ? `https://etro.gg/embed/gearset/${link}` : 
+                src:
+                  type === "etro" ? `https://etro.gg/embed/gearset/${link}` :
                   link,
                 class: "w-full h-full",
               }))
```
amakiirs, line 40:
```
// Normalize links where applicable
switch(type) {
  case "etro":
    const etroLink = link.match(/etro.gg\/gearset\/([A-Za-z0-9-]+)(?:[?#]|$)/i);
    if(!etroLink) {
      link = `https://etro.gg/embed/gearset/${link}`
    }
     break;
  default:
    break;
}

// Handle special cases of the BiS display
switch(type) {
  case "plain-text":
    // Plain text does not need an iframe, and can just be included as the given text (in the link field)
    var bisFrame = link;
    break;
  default:
    var bisFrame = h("div", { class: "h-96" }, h("iframe", {
      src: link,
      class: "w-full h-full"
    }));
    break;
}
```

Please forgive any possible formatting issues --- this is done on mobile.

Something like this makes a bit clearer what's being done and also ensures that we're always making decisions based on the type, rather than the contents of the link field. This code can replace the marked region (plus line 41 afterwards), and I believe it should have the same effect (though I may have incorrectly copied things from the PR)

static/theme-assets/editor/job-preview.js
```diff
@@ -22,17 +22,33 @@ const renderBisList = function (bis) {
...
-              }))
-              : link;
+        let link = bis.link;
+        
+        switch(type) {  
+          case "etro":  
+            const etroLink = link.match(/etro.gg\/gearset\/([A-Za-z0-9-]+)(?:[?#]|$)/i);  
+            if(!etroLink) {  
+              link = `https://etro.gg/embed/gearset/${link}`  
+            }
```
amakiirs:
```suggestion
            if(etroLink) {
              link = etroLink[1]
            }
            // Make etro links always become the embedded form
            link = `https://etro.gg/embed/gearset/${link}`
```

Sorry, misread this behaviour slightly in the last suggestion.

lailleo (approved):

PR #418, opened 2025-08-22
add validation for xivgear embed, add case for etro iframe generation, make xivgear field consistent with prod

kromuluss:
1.  forces the user to use an embed link when the type is set to xivgear to make sure the correct link is being input
image

2. merges the etro case into the primary switch call to generate its own custom iframe using the same class the site uses in prod
image

3. changes every field calling "xivgearset" in dev.yml to "xivgear" to match the prod config.yml file so that we don't have to continue doing checks using an outdated field name and then correct them before shipping to live

4. changes the default case to use the h-96 class to be consistent with prod's embed height for things like gsheets or generic iframes, and adds a validation check to the default case to make sure the user isn't placing an xivgear or etro link in a type that is not xivgear/etro and also isn't plaintext/genericlink
image

image
image

### DIFF
--- a/exampleSite/static/admin/dev.yml
+++ b/exampleSite/static/admin/dev.yml
@@ -1459,7 +1459,7 @@ collections:
         - sleepyshiba
         - genericiframe
         - genericlink
-        - xivgearset
+        - xivgear
         required: false
         widget: select
       - label: Link
@@ -2228,7 +2228,7 @@ collections:
         - sleepyshiba
         - genericiframe
         - genericlink
-        - xivgearset
+        - xivgear
         required: false
         widget: select
       - label: Link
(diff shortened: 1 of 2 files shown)


xpdota:
I don't know if you're able to use async calls in this context, but you can use the `/validateEmbed` endpoint to validate that it is actually a valid embed.

Valid: https://api.xivgear.app/validateEmbed?page=embed|bis|sge|endwalker|anabaseios&onlySetIndex=2
Not valid: https://api.xivgear.app/validateEmbed?page=bis|sge|endwalker|anabaseios&onlySetIndex=2
Also not valid: https://api.xivgear.app/validateEmbed?page=embed|bis|sge|endwalker|anabaseios
Complete nonsense links will give an error though, I can work on improving that if this is implemented: https://api.xivgear.app/validateEmbed?page=embed|bis|aaa|bbb|ccc&onlySetIndex=2

Obvious this isn't required for this round of changes, but something to keep in mind.

PR #461, opened 2025-11-19
Improve the modal dialogs

xpdota:
Uses a real `<dialog>` element instead of a purely JS-based solution


kromuluss (approved):
works on my machine trademark

PR #476, opened 2026-08-23
fix mathjax loading, triggering it after page load

VioletStardustXIV:
fixes mathjax loading to retrigger after page load, which lets stuff load asynchronously/optimally but means the latex renders properly

background: the latex rendering seemed to somehow become buggy and require a reload to work, reproduced it locally and this seems to fix it

### DIFF
--- a/layouts/partials/math.html
+++ b/layouts/partials/math.html
@@ -1,4 +1,3 @@
-<script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-chtml.js"></script>
 <script>
   MathJax = {
     tex: {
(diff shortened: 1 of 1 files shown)


layouts/partials/math.html
```diff
@@ -1,9 +1,21 @@
...
\ No newline at end of file
+
+    function triggerMathJax() {
+    if (window.MathJax && window.MathJax.typesetPromise) {
+      window.MathJax.typesetPromise();
+    }
+  }
+
+  document.addEventListener("DOMContentLoaded", triggerMathJax); // Normal page load
+  document.addEventListener("swup:contentReplaced", triggerMathJax); // Swup engine
```
xpdota, line 16: I don't think swup, turbolinks, or turbo is in use here

kromuluss (approved):

